PR #14, opened 2020-10-14
adding my name

Mila-ashtari:
added my name to .env file


server/.env
```diff
@@ -1,2 +1,3 @@
 TEAM_NAME="Shums"
+TEAM_NAME="Mila"
```
aeciorc, line 2: You want to append it to the existing variable :). You can see the value on dev is now:
`TEAM_NAME="Kevin, Adams, Sam`. So add yours to the end of it

PR #17, opened 2020-10-15
Feature Backend Signup Login

holley2032:
--What it does--
Creates a backend register and login API. Also adds a middleware for protected routes.

--What it avoids--
I spent hours trying to figure out how to get the imports to work and it still isn't perfect. If someone who knows more about Python imports wants to correct what I've done, that would be great!
As a result of that, testing looks like more of a pain than is worth trying to do right now, I think it's better to wait and hear feedback first, try to fix what's wrong with the imports, and then try to make sure it works. I don't believe that this is the final version of the API by any means, but it does get something on the table to work with.

--What to review--
All of it probably! Security, style, functionality, efficiency, I really wasn't sure what to do with this issue, so I tried cobbling together what made sense, but I could be on the wrong track altogether!

--Side note--
I added Mila to this review so she could see what I'm doing on the backend so she can align her coding with it. The main things to note are that the APIs are named RegisterAPI, LoginAPI, and UserAPI, all located in server/api/auth_handler. The RegisterAPI needs a JSON file with the fields "email", "password", and "confirm" [for the confirm passowrd field], and the LoginAPI needs a JSON file with the fields "email" and "password". If you already have different fields/names in effect, I can change them on the backend too if that's best.


server/models/models.py
```diff
@@ -0,0 +1,42 @@
+from server import db, flask_bcrypt, app
+import datetime
+import jwt
+
+
+class User(db.Model):
```
ThinkTink: Let's name this file user.py; we'll probably have more models eventually and will want to separate them. Then it'd be from models.user import User above in auth_handler

server/api/auth_handler.py
```diff
@@ -0,0 +1,152 @@
...
+                    "status": "fail",
+                    "message": "Some error occurred. Please try again."
+                }
+                return make_response(jsonify(responseObject)), 401
+        else:
+            responseObject = {
+                "status": "fail",
+                "message": "User already exists. Please log in."
+            }
+            return make_response(jsonify(responseObject)), 202
```
ThinkTink: I think 409 would be a more appropriate response code here

server/api/auth_handler.py
```diff
@@ -0,0 +1,152 @@
...
+                    responseObject = {
+                        "status": "fail",
+                        "message": "Password must be at least six characters long."
+                    }
+                    return make_response(jsonify(responseObject)), 401
+                if post_data.get("password") != post_data.get("confirm"):
+                    responseObject = {
+                        "status": "fail",
+                        "message": "Passwords must match."
+                    }
```
ThinkTink, line 54: need a return here as well

server/api/auth_handler.py
```diff
@@ -0,0 +1,152 @@
...
+                    return make_response(jsonify(responseObject)), 200
+                else:
+                    responseObject = {
+                        "status": "fail",
+                        "message": "User does not exist."
+                    }
+                    return make_response(jsonify(responseObject)), 404
+        except Exception as e:
+            print(e)
```
ThinkTink: No longer need this print statement, right?

server/api/auth_handler.py
```diff
@@ -0,0 +1,152 @@
...
+            responseObject = {
+                "status": "fail",
+                "message": "Try again"
+            }
+            return make_response(jsonify(responseObject)), 500
+
```
ThinkTink, line 128: Can we have more specific error messages based on 1)if user does not exist and 2) password does not match the user?

server/config.py
```diff
@@ -1,3 +1,21 @@
...
 TEAM_NAME = os.environ['TEAM_NAME']
+POSTGRES_USER = os.environ['POSTGRES_USER']
+POSTGRES_PW = os.environ['POSTGRES_PW']
+POSTGRES_URL = os.environ['POSTGRES_URL']
+POSTGRES_DB = os.environ['POSTGRES_DB']
+POSTGRES_LOCAL_BASE = 'postgresql://postgres@localhost/'
+
+
+class BaseConfig:
+    SECRET_KEY = os.getenv("SECRET_KEY", "example_backup_secret_key")
```
ThinkTink, line 15: should 'SECRET_KEY' be added to the .env file?

server/models/models.py
```diff
@@ -0,0 +1,42 @@
...
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    email = db.Column(db.String(255), unique=True, nullable=False)
+    password = db.Column(db.String(255), nullable=False)
+    registered_time = db.Column(db.DateTime, nullable=False)
+
+    def __init__(self, email, password, registered_time):
+        self.email = email
+        self.password = flask_bcrypt.generate_password_hash(password, app.config.get("BCRYPT_LOG_ROUNDS")).decode()
+        self.registered_time = registered_time
```
ThinkTink: It doesn't look like you're passing registered_time when you create a user in RegisterAPI. Should the time be set here in __init__ instead?

ThinkTink (commented):
Left some initial thoughts on this to consider

holley2032:
@ThinkTink I have addressed your comments in my latest commit, and I also added cookies (I think!) to the APIs. Thank you for your comments, they have been helpful!

server/api/auth_handler.py
```diff
@@ -0,0 +1,168 @@
...
+                    password=post_data.get("password")
+                )
+                db.session.add(user)
+                db.session.commit()
+                auth_token = user.encode_auth_token(user.id)
+                g.user = user
+                responseObject = {
+                    "status": "success",
+                    "message": "Successfully registered.",
+                    "auth_token": auth_token.decode()
```
aeciorc: Leave the auth_token out of the response object, since it's already going to be in the cookie

server/api/auth_handler.py
```diff
@@ -0,0 +1,168 @@
...
+                }
+                resp = make_response(jsonify(responseObject))
+                resp.set_cookie("Authentication token", auth_token.decode(), httponly=True)  # Should the auth_token be encoded or decoded?
```
aeciorc: The auth_token should be encoded. It will sit in the client's browser and will be passed automatically on any request they make. On those requests, your auth decorator will decode before allowing access to a protected route.

server/api/auth_handler.py
```diff
@@ -0,0 +1,168 @@
...
+                auth_token = user.encode_auth_token(user.id)
+                if auth_token:
+                    g.user = user
+                    responseObject = {
+                        "status": "success",
+                        "message": "Successfully logged in",
+                        "auth_token": auth_token.decode()
+                    }
+                    resp = make_response(jsonify(responseObject))
+                    resp.set_cookie("Authentication token", auth_token.decode(), httponly=True)  # Should the auth_token be encoded or decoded?
```
aeciorc: same comment as above

server/api/auth_handler.py
```diff
@@ -0,0 +1,168 @@
...
+            ).first()
+            if user and flask_bcrypt.check_password_hash(
+                user.password, post_data.get("password")
+            ):
+                auth_token = user.encode_auth_token(user.id)
+                if auth_token:
+                    g.user = user
+                    responseObject = {
+                        "status": "success",
+                        "message": "Successfully logged in",
```
aeciorc: same as above

server/api/auth_handler.py
```diff
@@ -0,0 +1,168 @@
...
+
+class UserAPI(MethodView):
+    def get(self):
+        auth_header = request.headers.get("Authorization")
+        if auth_header:
+            auth_token = auth_header.split(" ")[1]
+        else:
+            auth_token = ""
+        if auth_token:
+            resp = User.decode_auth_token(auth_token)
```
aeciorc: Here you would actually check the request.cookies (you'd look for a cookie with the same name as the one you set) rather than the headers. But let's not do this inside each route as that will get repetitive. Instead, use a decorator that does that and can be reused across requests.

server/api/auth_handler.py
```diff
@@ -0,0 +1,168 @@
...
+            return make_response(jsonify(responseObject)), 401
+
+
+# Basic middleware for protected routes pulled from Flask's documentation.
+def login_required(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        if g.user is None:
+            return redirect(url_for('login', next=request.url))
+        return f(*args, **kwargs)
```
aeciorc: Here you  check if the user is in the "g" object. Instead, you'll want to check if the tokens in the cookie is valid. Note that the value of "g" is wiped between requests, it's only used for storing data temporarily within the context of a single request. You can use it to add the user object after you validate the token, though. You can take the user id, find the user and add it to g. Then it can be used in the route

aeciorc (changes requested):

holley2032:
@aeciorc I've addressed your comments in the latest commit, and I tested the APIs using Postman. It largely seems to be functioning as intended.

server/api/auth_handler.py
```diff
@@ -7,6 +7,24 @@
...
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        try:
+            auth_header = request.cookies.get("Authentication token")
+            if auth_header:
+                auth_token = auth_header
+                auth_token = User.decode_auth_token(auth_token)
+                if not isinstance(auth_token, str):
+                    g.user = User.query.filter_by(id=auth_token).first()
+        except:
```
aeciorc, line 25: In this case, let's handle this here in the decorator, rather than in the calling function. Return an error message to the client and do not call the wrapped function

server/api/auth_handler.py
```diff
@@ -105,16 +120,11 @@ def post(self):
...
-        else:
-            auth_token = ""
-        if auth_token:
-            resp = User.decode_auth_token(auth_token)
-            if not isinstance(resp, str):
-                user = User.query.filter_by(id=resp).first()
+        try:
+            if g.user:
+                user = g.user
                 responseObject = {
```
aeciorc: Once you make the change above, you won't need to handle this here anymore

server/models/user.py
```diff
@@ -4,6 +4,8 @@
 
 
 class User(db.Model):
+    # I'm not entirely certain how to get these commands to create the table for us, I had to create the table
```
aeciorc: db.create_all() https://flask-sqlalchemy.palletsprojects.com/en/2.x/quickstart/

server/api/auth_handler.py
```diff
@@ -18,8 +18,24 @@ def decorated_function(*args, **kwargs):
...
+                        "status": "fail",
+                        "message": "User is not logged in."
+                    }
+                    return make_response(jsonify(responseObject)), 401
+            else:
+                responseObject = {
+                    "status": "fail",
+                    "message": "User is not logged in."
+                }
+                return make_response(jsonify(responseObject)), 401
```
aeciorc: To avoid repetition, you can raise an exception instead, which will be caught below.

aeciorc (approved):
One minor change and you're good to merge this!

server/api/auth_handler.py
holley2032: I fixed the issue by raising a generic exception, is that okay or do I need to raise something more specific?

aeciorc: Good for now 👍

PR #21, opened 2020-10-16
Login signup pages

Mila-ashtari:
--What it does--
Login and signup routes are created, and the user inputs are recorded and should be sent to the backend. styling still needs some work.

--What to review--
labels of the input fields are not positioned correctly


client/src/components/Login.js
```diff
@@ -0,0 +1,72 @@
...
+            '& p':{
+             margin:0,
+             marginRight:'10px'
+            }
+        }
+    },
+});
+function Login (){
+    const classes = useStyles();
+    const [userName, setName]=useState('')
```
ThinkTink: We don't need this variable, right?

client/src/components/Login.js
```diff
@@ -0,0 +1,72 @@
...
+    const [userEmail,setEmail]=useState('')
+    const [userPass, setPass]=useState('')
+    return(
+        <section className={classes.root}>
+            <form >
+                <h2>Login</h2>
+                <label>Your email address:</label>
+                <TextField className='textField' id="outlined-basic" variant="outlined" label="email" placeholder="email" required type="email"  onChange={(e)=>setEmail(e.target.value)}/>
```
ThinkTink: A couple of notes here:
- id should be unique, and this has the same id as the one below
- We probably want value={userEmail} to properly link the textfield to the input variable.

client/src/components/SignUp.js
```diff
@@ -0,0 +1,82 @@
...
+                    <Button variant="contained" color="secondary">Create an account</Button>
+                </form>
+                <div className='login'>
+                    <p>Already have an account?</p>
+                    <Link  className='loginLink' to="/login">login</Link>
+                </div>
+            </div>
+        </section>
+    )
+}
```
ThinkTink, line 80: I'm not sure what exactly is the causing the label alignment problem, but here's one thing to try:
https://stackoverflow.com/questions/46966413/how-to-style-material-ui-textfield discussess using InputProps
it looks like there's an InputLabelProps to: https://material-ui.com/api/text-field/
So maybe try adding a class to that and see if it fixes the problem

client/src/App.css
```diff
@@ -1,33 +1,40 @@
...
+html {
+    box-sizing: border-box;
+  }
+  
+  html, body, #root {
+    padding: 0px !important;;
+    margin: 0px !important;;
+    width: 100vw;
+  }
+
```
ThinkTink: Generally we want to avoid using !important in CSS if possible.

Also, is the unformatted section above a CSS reset? If so, let's mark it with a comment explaining it. Also, if so, is it necessary? Ideally we would just have minimal styling on App and it would be using makeStyles to be consistent with the rest of the app

client/src/components/Login.js
```diff
@@ -0,0 +1,72 @@
...
+        borderRadius:'2%',
+        width:'50% ',
+        margin: '50px auto',
+        textAlign:'center',
+        backgroundColor:'rgb(241, 238, 232)',
+        padding:'30px',
+        '& form':{
+            '& h2':{
+            },
+            '& .textField':{
```
ThinkTink: I think it's more clear to have each new class name be on its own. Like:
root: {
   border, etc...
},
textField: {
   width, etc...
}

client/src/components/SignUp.js
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+});
+function SignUp (){
+    const classes = useStyles();
+    return(
+        <section className={classes.root}>
+            <div className='container'>
+                <form >
+                    <h2>Sign Up</h2>
```
ThinkTink: Material UI has a Typography that would be good for here: https://material-ui.com/components/typography/

client/src/components/SignUp.js
```diff
@@ -0,0 +1,82 @@
+import React, { useState, useEffect } from 'react';
+import { BrowserRouter, Route, Link, Redirect, Switch } from "react-router-dom";
```
ThinkTink: We only need to import what we're using in this file specifically

client/src/components/SignUp.js
```diff
@@ -0,0 +1,82 @@
...
+import { makeStyles } from '@material-ui/core/styles';
+import TextField from '@material-ui/core/TextField';
+import Button from '@material-ui/core/Button';
```
ThinkTink: We can simplify imports to one line here, see: https://material-ui.com/guides/minimizing-bundle-size/

client/src/components/Login.js
Mila-ashtari: are you refering to classes?

client/src/components/Login.js
Mila-ashtari: got it

client/src/components/Login.js
ThinkTink: I'm referring to userName

client/src/components/SignUp.js
```diff
@@ -0,0 +1,37 @@
...
+
+
+const useStyles = makeStyles({
+    root: root
+});
+function SignUp (){
+    const classes = useStyles();
+    return(
+        <section className={classes.root}>
+            <div className='container'>
```
morgan-sam: For this container I would change the following:

1. Change `div` to Material UI's `Box`
2. Change `className` from `'container'` to `{classes.container}`
3. Add the container class to useStyles:
```
const useStyles = makeStyles({    
    container : {
        border: 'solid rgb(241, 238, 232) 1px',
        borderRadius: '1%',
        width: '50%',
        maxWidth: '400px',
        minWidth: '250px',
        margin: '0 auto',
        textAlign: 'center',
        backgroundColor: '#f7f3f3',
        padding: '30px',
}); 
```
4. Replace the padding with `theme.spacing`:
```
        ...
        backgroundColor: '#f7f3f3',
        padding: theme.spacing(1)
}); 
```

client/src/style/signupLoginPages.js
```diff
@@ -0,0 +1,55 @@
...
+        padding:'20px',
+        borderTop:'1px solid rgba(128, 128, 128, 0.274)',
+        justifyContent:'center',
+        marginTop:'40px',
+        textAlign:'center',
+        '& p':{
+         margin:0,
+         marginRight:'10px'
+        },
+        '& .loginLink, & .signUpLink':{
```
morgan-sam: These could be a single class inside useStyles, as they are both the same.
```
const useStyles = makeStyles({    
    loginSignupLink : { color: '#f50057' }
}); 
```

Then you can just use the same class for each different link:

`<Link  className={classes.loginSignupLink} to="/login">login</Link>`

`<Link className={classes.loginSignupLink} to="/">Create an account</Link>`

client/src/components/SignUp.js
```diff
@@ -0,0 +1,37 @@
...
+
+import root from '../style/signupLoginPages'
+
+
+const useStyles = makeStyles({
+    root: root
+});
+function SignUp (){
+    const classes = useStyles();
+    return(
```
aeciorc, line 60: You're missing the state and event handling in this component. Once you add that, it should be good to merge

aeciorc (changes requested):
Just one comment above

client/src/style/signupLoginPages.js
morgan-sam: (edited example as I made a couple mistakes)

PR #22, opened 2020-10-16
PR: Setup for products and list tables, and testing routes using api requests

### LINKED ISSUE #7: BE: Lists, products
- Product model has description, name, url and price. 
-List  model has several products, does not need a cover image yet
-Routes to create and edit lists and product

adams-liu:
**What it does:**
 - Created the products and list tables in the database
 - Created routes to add and edit products and lists

**What to look for:**
- Folder structure, don't know what the best way to organize my functions, routes, db instances, etc. 
- `Line 17: from models import List, Product` I'm importing from models.py and models.py is importing from app (probably very bad practice)
- I've cover 5 API requests, I don't know if there are any more I am missing. 
- Also, I don't know if I am doing error handling correctly, I did a general try/except statement

**What to avoid**
- I'm not using blueprints atm, but I've had the code setup just to test the the api requests, and routing.  I will make sure to use blueprints as soon as the folder structure is in place.

Thank you!


ThinkTink:
Some thoughts on file organization:
- similar to how Austin has a server/models/user.py, you may want to create a server/models/database.py. I'm not sure the best name, but basically a file within a models folder that contains the List and Product models. Then you can have `from models.database import List, Product`
- also to keep in sync with Austin's work, the routes could live in a server/api/database_handler.py
- as for managing the database instances, we wouldn't generally want to create tables every time we run the code because we want our data to persist rather than creating a new table every time. So ideally we create the database outside of running the app and then just update the existing database via api calls from the front end as users interact with the site.

server/app.py
```diff
@@ -12,4 +12,77 @@
...
+    
+    if request.method == "DELETE":
+        body = json.loads(request.get_data())
+        product_id = body['product_id']
+        product_name = Product.query.filter_by(id=product_id).first().name
+        Product.query.filter_by(id=product_id).delete()
+        db.session.commit()
+        return jsonify({'response': "{} was successfully deleted from the database".format(product_name)}), 200
+
+
```
ThinkTink, line 30: I think the routes and their purposes are not very intuitive (like a /edit that doesn't take a PUT or POST, and GET to /add_product returns lists not products?).
Thinking through what all we need, from the front end:
1) a GET request that will allow us to display all of the lists in a user's account. So takes a user id and gets their lists?
[->so we need to associate user ids with lists]
2) a GET request that will allow us to show all the products on a list. Alternatively, we could just fetch all the data from 1 and 2 at the same time from the same route
3) a POST request to add a list
4) a POST request to add a product to a list
5) a DELETE request to delete a product from a list
6) a DELETE request to delete a list (though I don't think this is actually in the mock up, but it would make sense?)

So we could simplify this to
/lists - GET for get by user id, POST for add, DELETE for delete
/products - GET for get by list id, POST for add, DELETE for delete

You don't need to do it that way, but do take a look at the organization and route naming to making sure it makes sense, and maybe add some explanatory comments

server/app.py
```diff
@@ -12,4 +12,77 @@
...
+def createList():
+    if request.method == 'POST':
+        body = json.loads(request.get_data())
+        list_item = List(body['product_type'])
+        db.create_all()
+        db.session.add(list_item)
+        try:
+            db.session.commit()
+        except:
+            return jsonify({'response': "{} already exists in the database".format(body['product_type'])}), 400
```
ThinkTink: There's other reasons this could fail, right? Maybe we can except Exception as e: and return e.message

server/app.py
```diff
@@ -12,4 +12,77 @@
...
+    return jsonify({'response': "{} was successfully added to the database".format(body['product_type'])}), 200
+
+
+@app.route('/add_product', methods=['GET', 'POST'])
+def addProduct():
+    if request.method == "GET":
+        product_types = List.query.all()
+        ls = []
```
ThinkTink: let's use a more descriptive variable name, and same with the ls below

server/app.py
```diff
@@ -12,4 +12,77 @@
...
+                {
+                    "id": item.id,
+                    "product_type": item.product_type
+                })
+        return jsonify({'list': ls}), 200
+
+    elif request.method == 'POST':
+        body = json.loads(request.get_data())
+        product_item = Product(
+            body['list_id'], body['name'], body['description'], body['price'], body['url'])
```
ThinkTink: Let's add some error checking - what if the incoming request doesn't have all of these fields?
Also, maybe not in this PR, but we may want to think about adding data validation, like price needs to be a float, etc

server/app.py
```diff
@@ -12,4 +12,77 @@
...
+        for prod in products:
+            ls.append(
+                {
+                    "id": prod.id,
+                    "name": prod.name,
+                    "description": prod.description,
+                    "price": prod.price,
+                    "url": prod.url
+                })
+        return jsonify({product_type.product_type: ls}), 200
```
ThinkTink: Maybe change the name of product_type to list, so we can say list.product_type here

server/models.py
```diff
@@ -0,0 +1,47 @@
+from app import db
+
+
+class List(db.Model):
+
+    __tablename__ = 'list'
+
+    id = db.Column(db.Integer, primary_key=True)
```
ThinkTink: Let's add autoincrement here and on the id for Product

server/app.py
adams-liu: I tried to create the routes based on the 3 provided views given in the [google drive](https://drive.google.com/drive/folders/1_-AKG8TNKAqSrK9E7cs_Z5gLcQYtnMny) because some views call for different types of request (eg. create_list you need to get a list of all product_types as well as be able to post)

server/app.py
adams-liu: e.message was deprecated, instead I used `e.__cause__`, hope that's alright

server/app.py
adams-liu: so by default all the input is of type string, I only did validation for ` body['price']` because this requires a 2-decimal float. `body['list_id']` I casted it to type int().

server/app.py
adams-liu: so I was thinking there was 3 different routes each representing a view

adams-liu:
my `app.py` now looks like this:

```
from flask import Flask
from flask_sqlalchemy import SQLAlchemy
from flask_migrate import Migrate


app = Flask(__name__)
app.config.from_object('config.Config')

db = SQLAlchemy(app)
migrate = Migrate(app, db)


from api.ping_handler import ping_handler
from api.home_handler import home_handler
from api.database_handler import database_handler

app.register_blueprint(home_handler)
app.register_blueprint(ping_handler)
app.register_blueprint(database_handler)
```
the problem is my database_handler.py requires the SQLAlchemy driver that is initialized in app.py so, I'm only able to import it after I created my  `db = SQLAlchemy(app)` how can I solve this issue?

adams-liu:
@aeciorc what do you think?

server/app.py
adams-liu: @aeciorc How do you think I should best set-up the routing. I set them up by the different views of the dashboard, but I assume it wouldn't matter because the dialog views don't actually change the webpages

adams-liu:
@aeciorc what do you think is the best way to set-up SQLAlchemy instance?

aeciorc:
> @aeciorc what do you think is the best way to set-up SQLAlchemy instance?

Use the app factory pattern:
```python
from flask import Flask
from flask_sqlalchemy import SQLAlchemy
from flask_migrate import Migrate
from database import db
from api.ping_handler import ping_handler
from api.home_handler import home_handler
from api.database_handler import database_handler

migrate = Migrate()


def create_app():
    app = Flask(__name__)
    app.config.from_object('config.Config')

    print('init')
    db.init_app(app)


    migrate.init_app(app, db)

    with app.app_context():


        # Register Blueprints
        app.register_blueprint(home_handler)
        app.register_blueprint(ping_handler)
        app.register_blueprint(database_handler)


        db.create_all()
        return app


app = create_app()


```

note that I declare db in a different file, but initialize here. This is the file: 
```python
from flask_sqlalchemy import SQLAlchemy

db = SQLAlchemy()
```

Anytime you need to import the db (ie from database_handler.py or the /models) use this file instead


source:https://hackersandslackers.com/flask-application-factory/

aeciorc (commented):
Let's also refactor:
- The routes (as per the comment on app.py)
- The file structure for handlers. Rather than a database_handler, let's have separate handlers for different entities, like lists_handler, products_handler, etc
- Models, let's have one file per entity. So models/product.py, models/list.py, etc

aeciorc (approved):
Alright, looks good! I'm approving it, but here's one more change I'd like:
For routes, the name of a resource should be plural. So /lists, /products, etc. See some info on the rationale for that here https://restfulapi.net/resource-naming/

PR #23, opened 2020-10-16
Feature dashboard skeleton

morgan-sam:
FE: Dashboard Skeleton #8 

<br />

**What it does:**

Skeleton of the app dashboard. Includes the Header & Shopping Lists page.

Introduced `@material-ui/icons` dependency.

Currently uses demo images/text to display the lists. Will be replaced with actual API calls.

Deleted the starter Landing & Ping pages.

<br />

**What to look for:**

- Styling:

  - Some repeating styles, needs to be centralized

  - All margin/padding values should be converted to `theme.spacing(n)
`
- UI: 

  - Visual spacing of app could be improved

  - Paste link separator not full container height

  - Tiny gap in between notifications arrow and border

  - Plus sign IconButton large and very close to text

- Coding style:

  -  Tab/Button text made bold using `textShadow: "0 0 0.01px black"` (`fontWeight: "bold"` caused the text to move) 

  -  Some inline styling on components (i.e. [TabIndicatorProps](https://github.com/hatchways/team-apple-cider/compare/feature-dashboard-skeleton?expand=1#diff-e254af1e315da8bf95150461ac1f09e8cc9b11cf5e408436efeb17e18441a3aaR52))

  -  State of Dashboard passed from top parent using props (Context may be preferable)

  -  The Notification tab is actually a Button, as Tab caused complications: 

     - NotificationPopup would not position correctly inside a Tab
     - Clicking the notification tab would also change the selectedPage to blank

<br />

Also just to note:  Some commits were only tiny changes to UI, but I can squash those before merging if needed.

<br />

**What to avoid:**

- No logout functionality

- No use of User Route

- Parent component does not hold User Info


client/src/components/AddItem.js
```diff
@@ -0,0 +1,92 @@
...
+          disableUnderline
+        >
+          <MenuItem value="none" disabled>
+            Select List
+          </MenuItem>
+          {demoListArray.map((el, i) => (
+            <MenuItem key={i} value={el}>
+              {el}
+            </MenuItem>
+          ))}
```
ThinkTink: Let's use more descriptive variable names inside the map. Eventually we'll have a list id or something to use for the key, but index is fine for now.

client/src/components/HeaderTabs.js
```diff
@@ -0,0 +1,67 @@
...
+  const handleTabChange = (e, newValue) => setSelectedPage(newValue);
+
+  return (
+    <Box className={classes.tabContainer}>
+      <Tabs
+        value={selectedPage}
+        onChange={handleTabChange}
+        aria-label="page select"
+        TabIndicatorProps={{
+          style: { backgroundColor: "white" },
```
ThinkTink: let's change this to using a className

client/src/components/ListsDisplay.js
```diff
@@ -0,0 +1,111 @@
...
+  const classes = useStyles();
+
+  return (
+    <Box className={classes.shoppingContainer}>
+      <Typography variant="h5" className={classes.listsTitle}>
+        My Shopping Lists:
+      </Typography>
+      <Box className={classes.myShoppingLists}>
+        {demoList.map((el, i) => (
+          <Box className={classes.listContainer} key={i}>
```
ThinkTink: More descriptive variable names here too

client/src/components/NotificationPopup.js
```diff
@@ -0,0 +1,53 @@
...
+    top: "0",
+    transform: "translate(-50%,-20%)",
+    display: "flex",
+    flexDirection: "column",
+    alignItems: "center",
+    margin: "4rem",
+    position: "absolute",
+    width: "12rem",
+  },
+
```
ThinkTink: No need for extra spacing between these

client/src/components/NotificationsButton.js
```diff
@@ -0,0 +1,61 @@
...
+          disableRipple
+          onClick={() => setNotificationsOpen((cur) => !cur)}
+        >
+          <Typography
+            className={classes.buttonText}
+            style={{
+              textShadow: Boolean(notificationsOpen)
+                ? "0 0 0.01px black"
+                : "none",
+            }}
```
ThinkTink: Can we conditionally add a second className instead?

client/src/components/ProfileMenu.js
```diff
@@ -0,0 +1,50 @@
...
+  const [anchorEl, setAnchorEl] = useState(null);
+  const classes = useStyles();
+  const handleClick = (e) => setAnchorEl(e.currentTarget);
+  const handleClose = () => setAnchorEl(null);
+
+  return (
+    <Box className={classes.profileMenuContainer}>
+      <img className={classes.profileIcon} src={icon} alt="profile" />
+      <Button onClick={handleClick}>
+        <Typography style={{ textTransform: "none" }}>Profile</Typography>
```
ThinkTink: let's use a className here as well

client/src/pages/Dashboard.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+  bodyContainer: {
+    flex: "1",
+    backgroundColor: "#f8f8f8",
+  },
+}));
+
+const Dashboard = () => {
+  const [selectedPage, setSelectedPage] = useState(0);
+  const [notificationsOpen, setNotificationsOpen] = useState(0);
```
ThinkTink: This should probably start as false, not 0

client/src/pages/Friends.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { Box, Typography } from "@material-ui/core";
+
```
ThinkTink, line 16: Let's convert this component to makeStyles

client/src/components/NotificationsButton.js
```diff
@@ -0,0 +1,61 @@
...
+          </Typography>
+          <Box className={classes.notificationIcon} />
+        </Button>
+        {Boolean(notificationsOpen) && <NotificationPopup {...props} />}
+      </Box>
+    </ClickAwayListener>
+  );
+};
+
+export default NotificationsButton;
```
ThinkTink, line 65: I think making this a Button is fine

ThinkTink (commented):
This is looking good! Go ahead and make the changes you mentioned in the PR, but as we discussed this morning don't worry the notifications right now. We should discuss as a group if using Context is a good approach, so leave as prop passing for now I think.

client/src/components/HeaderTabs.js
morgan-sam: Would [this](https://github.com/hatchways/team-apple-cider/pull/23/commits/ba7bb4de0dbee530f9f9628836b0119800dd2067) suffice, or does it need to be using className specifically?

ThinkTink: Yup, that looks fine!

Issue #6, opened 2020-10-13
BE: Amazon Scraping Service


aeciorc:
-A service that takes a URL and returns product details
-Use a simulated browser (selenium - webdriver)

PR #31, opened 2020-10-19
Feature Amazon Scraping Service

morgan-sam:
BE: Amazon Scraping Service #6

**What it does:**

Scrapes Amazon item URLs for their name, old price, current price, image URL, and availability. To use:
```
import scraper;
itemDetails = scraper.scrapeAmazon(input)
print(itemDetails)
```
Where input is equal to either a string URL or an array of string URLs.

Provided a bash script to run named `test_scraper.sh` with a few examples. 

**What to look for:**

- Any edge cases: I tested a variety of conditions; different country stores, out of stock items, in stock but only x amount left, unreleased items, etc. I accommodated all I could think of but if I missed any out, or if there are any unexpected results, let me know.
- Could be optimized to be faster. This falls under 2 main categories:
  - I have somewhat limited experience using async code in Python, so I'm aware their are likely better practices I'm missing. For example in JavaScript I would use for `Promises.all()` for asynchronously iterating over an array, whereas in my Python code when there was a list input I just ran each scrape synchronously.  
  - I'm not very familiar with the Selenium profiles/options. I disabled Flash/JavaScript and ran the browser in headless mode, this seemed to improve the speed somewhat. When I disabled images from loading, rather than giving me an image URL I would get a raw image string data (`data:image/png;base64,iVBORw0KGgoAAAA...`) but only with this [PS5 link](https://www.amazon.co.uk/dp/B08H95Y452).


server/scraper.py
```diff
@@ -0,0 +1,59 @@
...
+def scrapeAmazon(input):
+    try:
+        if (isinstance(input, list)):
+            items = []
+            for url in input: items.append(getItem(url))
+            return items
+        elif (isinstance(input, str)): return getItem(input)
+        else: return 'ERROR: INVALID INPUT'
+    finally:
+        browser.quit()
```
aeciorc: Notice that if you call scrapeAmazon 2x in a row, it will fail the 2nd time, because you quit the browser.

aeciorc: Let's wrap this as a class so that it can be instantiated more than once.

server/scraper.py
```diff
@@ -0,0 +1,59 @@
+import re
+from selenium import webdriver
+from selenium.webdriver.common.keys import Keys
+from selenium.webdriver.firefox.options import Options
+
+PATH = '/home/user/scraping/geckodriver'
```
aeciorc: Rather than manually installing the driver, let's use a library to do that for us: https://pypi.org/project/webdriver-manager/

aeciorc (commented):
Nice job, left a few comments below. 
Besides that, notice that a significant amount of time is actually spent initializing the browser... This is something we'll address later on, but basically we'll keep a single browser open (on production this would scale on demand) to service all scraping tasks. We'll push scraper jobs into our tasks queue, and one by one they'll be popped in handled by the scraper.

server/api/scraper.py
```diff
@@ -0,0 +1,49 @@
+import re
+from selenium import webdriver
+from selenium.webdriver.chrome.options import Options
+from webdriver_manager.chrome import ChromeDriverManager
+from webdriver_manager.utils import ChromeType
```
adams-liu, line 5: Make sure to do `pipenv install [LibraryName]` so that your pipfile is updated aswell (you might have only installed it the packages in your local environment)

PR #32, opened 2020-10-20
PR: Updating imports to work with auth_handler

adams-liu:
- “What it does”
     - I made structural changes to the `__init__.py` , `config.py`, and `app.py` 
     - The instances of the plugins are declared in `__init__.py` but initialized in app.py (aside from db)
     - I also updated the imports for your` auth_handler` and `user.py`

- “What to look for”
   - I've gotten my api handlers fully working, so now if you checkout to my branch and test your auth_handler hopefully is should work too.
   - Also double-check I didn't delete anything important.
   - I also changed your blueprint handler name to `auth_handler `instead of `auth_blueprint`, just so we have the same naming conventions

Thanks!


holley2032:
The authentication isn't working because the app is now initialized in app.py. Models.user.py requires the app to be imported so that it can retrieve SECRET_KEY and BCRYPT_LOG_ROUNDS from the config module. However, app.py imports auth_handler.py, which itself imports models.user.py. It's an import issue. I'm not sure how best to solve it, but the three solutions that come to mind for me are:

1. Move the app factory to __init__.py.
2. Move the blueprint registration to __init__.py.
3. Import SECRET_KEY and BCRYPT_LOG_ROUNDS directly from config.py. This has the disadvantage of not being able to maintain distinct production and development configurations.

There may be other options too, but these are what came to mind. Let me know if you want to talk about it!

adams-liu:
so I now rearranged some things I have it set so that you wouldn't call the auth_handler without initializing the app first

PR #33, opened 2020-10-20
Feature auth integration

holley2032:
--What it does--
-Integrates the front-end login and sign-up pages with the back-end APIs
-A touch of redesign to make the front-end work better with the added field of confirm password.
-Also adds support on the backend for the name variable in the form.

--What it avoids--
-I did not redirect users after registering or logging in. I think that would be better addressed once protected routes are finished.
-I did not notify the user what went wrong if the registration or login fails. I've largely kept this pull request to providing the basic functionality needed to make the front-end auth pages connect to the back-end auth APIs, but I want to acknowledge that there is more to do to make it operate the way we want it to.
-I did not store the name as individual first and last names in the database. That can be added if needed, but as is I think it may make more sense to avoiding adding a line to the form by asking for first name and last name.

--What to look for--
-There are a couple problems noted in the comments that I am unsure how to fix. It still works, but it could work better in some contexts.
-If there is an issue which I did not address that should be addressed in this pull request.

### DIFF
--- a/client/src/components/Login.js
+++ b/client/src/components/Login.js
@@ -59,15 +59,15 @@ function Login (){
     const classes = useStyles();
     const [email,setEmail]=useState('');
     const [password, setPassword]=useState('');
-    const [errors, setErrors]=useState('');
+    const [errors, setErrors]=useState({});
 
     const validations = () => {
-        let temp = {};
-        temp.email = email ? "" : "Please enter an email.";
-        temp.password = password ? "" : "Please enter a password.";
-        setErrors({ ...temp });
+        const errorsCopy = {...errors};
+        errorsCopy.email = email ? "" : "Please enter an email.";
+        errorsCopy.password = password ? "" : "Please enter a password.";
+        setErrors({ ...errorsCopy });
 
-        return Object.values(temp).every(x => x == "");
+        return Object.values(errorsCopy).every(field => field === "");
     }
 
     const handleClick = (e) => {
@@ -83,7 +83,7 @@ function Login (){
                     'password': password,
                 })})
                 .then(response => response.json())
-                .then(function(response) {
+                .then((response) => {
                     if (response.status === 'success') {
                         console.log('Success:', email);
                     }
@@ -97,16 +97,16 @@ function Login (){
                 })
             }
     }
-    // The console throws a warning because the 'error' field in each TextField is given a string, not a boolean. However, the code appears to work as intended, and it does not appear to be a problem.
+
     return(
             <section className={classes.login}>
                 <Box className={classes.formContainer}>
                     <form>
                         <h2 className={classes.h2}>Sign in</h2>
                         <label>Your email address:</label>
-                        <TextField className={classes.textField} variant="outlined" label="email"  fullWidth required type="email" error={errors.email} helperText={errors.email} onChange={(e)=>setEmail(e.target.value)}/>
+                        <TextField className={classes.textField} variant="outlined" label="email"  fullWidth required type="email" error={!!errors.email} helperText={errors.email} onChange={(e)=>setEmail(e.target.value)}/>
                         <label>Password:</label>
-                        <TextField className={classes.textField} variant="outlined" label="password" fullWidth required type="password" error={errors.password} helperText={errors.password} onChange={(e)=>setPassword(e.target.value)}/>
+                        <TextField className={classes.textField} variant="outlined" label="password" fullWidth required type="password" error={!!errors.password} helperText={errors.password} onChange={(e)=>setPassword(e.target.value)}/>
                         <Button className={classes.button} variant="contained" color="secondary" onClick={handleClick}>Login</Button>
                     </form>
                     <Box className={classes.signup}>
--- a/client/src/components/Login.js
+++ b/client/src/components/Login.js
@@ -57,8 +57,8 @@ signupLink:{
 }));
 function Login (){
     const classes = useStyles();
-    const [userEmail,setEmail]=useState('')
-    const [userPass, setPass]=useState('')
+    const [email,setEmail]=useState('')
+    const [password, setPass]=useState('')
     return(
             <section className={classes.login}>
                 <Box className={classes.formContainer}>
--- a/client/src/components/Login.js
+++ b/client/src/components/Login.js
@@ -59,6 +59,22 @@ function Login (){
     const classes = useStyles();
     const [email,setEmail]=useState('')
     const [password, setPass]=useState('')
+    const handleClick = () => {
+        fetch("/auth/login", {
+            method: 'POST',
+            headers: {
+                'Content-Type': 'application/json'
+            },
+            body: JSON.stringify({
+                'email': email,
+                'password': password,
+            })})
+            .then(response => response.json())
+            .then(console.log('Success')) // Logs success even when unsuccessful, not certain how to fix it.
+            .catch((error) => {
+                console.error('Error:', error)
+            })
+    }
     return(
             <section className={classes.login}>
                 <Box className={classes.formContainer}>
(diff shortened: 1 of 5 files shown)


client/src/components/Login.js
```diff
@@ -57,8 +57,24 @@ signupLink:{
 }));
 function Login (){
     const classes = useStyles();
-    const [userEmail,setEmail]=useState('')
-    const [userPass, setPass]=useState('')
+    const [email,setEmail]=useState('')
+    const [password, setPass]=useState('')
+    const handleClick = () => {
```
ThinkTink: A couple small cleanup notes:
- let's actually makes this setPassword to be consistent
- let's add semi colons 
- add a line break between the state and first function, for readability

client/src/components/Login.js
```diff
@@ -57,8 +57,24 @@ signupLink:{
...
+            method: 'POST',
+            headers: {
+                'Content-Type': 'application/json'
+            },
+            body: JSON.stringify({
+                'email': email,
+                'password': password,
+            })})
+            .then(response => response.json())
+            .then(console.log('Success')) // Logs success even when unsuccessful, not certain how to fix it.
```
ThinkTink: What we want to do here is look at the http status code that's in the response - if it's 200, we're good to go. if it's not, then we'd look at the message of the response to get the right error message to display

client/src/components/SignUp.js
```diff
@@ -54,9 +54,29 @@ signupLink:{
...
 function SignUp (){
     const classes = useStyles();
-    const [userName, setName]=useState('')
-    const [userEmail,setEmail]=useState('')
-    const [userPass, setPass]=useState('')
+    const [name, setName]=useState('')  // Changed names to correspond with backend.
+    const [email,setEmail]=useState('')
+    const [password, setPass]=useState('')
+    const [confirm, setConfirm]=useState('')
+    const handleClick = () => {
```
ThinkTink: Same small cleanup notes here as above

client/src/components/SignUp.js
```diff
@@ -54,9 +54,29 @@ signupLink:{
...
+                'password': password,
+                'confirm': confirm
+            })})
+            .then(response => response.json())
+            .then(console.log('Success:', name)) // Logs success even when unsuccessful, not certain how to fix it.
+            .catch((error) => {
+                console.error('Error:', error)
+            })
+    }
+    // Added confirm password field. Still needs to check for password being at least six characters
```
ThinkTink: We should also check here if the confirm password matches password. We can do these checks either on the input field (set onChange to a function that 1) sets the value and 2) validates the value) or on the submit (handleClick checks validation and either shows error messages or makes the API call). It's kind of a UX decision on which to use. Either way, we'll want to use the error and helperText on TextField: https://material-ui.com/components/text-fields/#validation

server/models/user.py
```diff
@@ -8,11 +8,15 @@ class User(db.Model):
 
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
     email = db.Column(db.String(255), unique=True, nullable=False)
+    name = db.Column(db.String(255), nullable=False)  # Right now this is a very simple full name column. Could be
+    # possibly better to include first and last name fields in the front-end form and make this two columns, one for
+    # first names and one for last names.
```
ThinkTink: Maybe. On the other hand, this way gives the user the freedom to use whatever they wish for their name, if they like to go by one name or 3+. I think it's fine for our use case.

ThinkTink (approved):
Looks good so far! Reading the response from the API and adding form validation shouldn't take too much to add.

Touch base with Mila about her progress on the protected routes - I think we may need to add something to local storage in your handleClick functions. Also, depending on how far along she is, we may want to just redirect to the dashboard on signup/login for now

ThinkTink:
Oops, I just meant to comment, not approve. Let me take a look at this again before you merge it!

client/src/components/SignUp.js
```diff
@@ -68,11 +88,13 @@ function SignUp (){
...
-                        <Button className={classes.button} variant="contained" color="secondary" >Login</Button>
+                        <label>Confirm Password:</label>
+                        <TextField className={classes.textField} variant="outlined" label="confirm password" fullWidth required type="password" onChange={(e)=>setConfirm(e.target.value)}/>
+                        <Button className={classes.button} variant="contained" color="secondary" onClick={handleClick}>Register</Button>
                     </form>
                     <Box className={classes.signup}>
-                        <p className={classes.p}>Don't have an account?</p>
-                        <Link className={classes.signupLink} to="/login">Create an account</Link>
+                        <p className={classes.p}>Already have an account?</p>
+                        <Link className={classes.signupLink} to="/login">Login</Link>
```
ThinkTink, line 121: Nice catch here!

client/src/components/SignUp.js
holley2032: It now updates the error and helperText when submit is clicked. I don't have a huge preference one way or another, but I do slightly prefer this way to avoid users being annoyed by being told their details are wrong as they are entering them.

holley2032:
@ThinkTink I have added in the changes you mentioned except for the protected route redirect, primarily to get the changes I've made so far pushed to this branch.

client/src/components/Login.js
```diff
@@ -57,18 +57,57 @@ signupLink:{
...
+                        window.alert(response.message);
+                        console.log(response.message);
+                    }
+                })
+                .catch((error) => {
+                    console.error('Error:', error)
+                })
+            }
+    }
+    // The console throws a warning because the 'error' field in each TextField is given a string, not a boolean. However, the code appears to work as intended, and it does not appear to be a problem.
```
ThinkTink: To fix this, you can coerce the string to a boolean: `error={!!errors.password}`

client/src/components/Login.js
```diff
@@ -57,18 +57,57 @@ signupLink:{
 }));
 function Login (){
     const classes = useStyles();
-    const [userEmail,setEmail]=useState('')
-    const [userPass, setPass]=useState('')
+    const [email,setEmail]=useState('');
+    const [password, setPassword]=useState('');
+    const [errors, setErrors]=useState('');
```
ThinkTink: If this eventually will be an object, we want to start it as an object - `=useState({})`

client/src/components/Login.js
```diff
@@ -57,18 +57,57 @@ signupLink:{
...
+
+    const validations = () => {
+        let temp = {};
```
ThinkTink: how about `const errorsCopy = {...errors}` ? We generally want to avoid vague variable names. Then below, `return Object.values(errorsCopy).every(field => field === "")`

client/src/components/Login.js
```diff
@@ -57,18 +57,57 @@ signupLink:{
...
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                body: JSON.stringify({
+                    'email': email,
+                    'password': password,
+                })})
+                .then(response => response.json())
+                .then(function(response) {
```
ThinkTink: let's use an arrow function here too

client/src/components/Login.js
```diff
@@ -57,18 +57,57 @@ signupLink:{
...
+                    if (response.status === 'success') {
+                        console.log('Success:', email);
+                    }
+                    else {
+                        window.alert(response.message);
+                        console.log(response.message);
```
ThinkTink, line 92: Not necessary for this PR, but eventually we'll want to use something like a snackbar for this kind of notification- https://material-ui.com/components/snackbars/

client/src/components/SignUp.js
```diff
@@ -7,7 +7,7 @@ const useStyles = makeStyles((theme)=>({
 login:{
     maxWidth:'1200px',
     margin:'0 auto',
-    height:'100vh',
+    height:'123vh', // Eyeballed to make the form centered in the background. May be a better way to do this.
```
ThinkTink: As long as it looks ok on a few different monitor sizes, this is fine - you can remove the comment

client/src/components/SignUp.js
```diff
@@ -54,25 +54,71 @@ signupLink:{
 }));
 function SignUp (){
     const classes = useStyles();
-    const [userName, setName]=useState('')
-    const [userEmail,setEmail]=useState('')
-    const [userPass, setPass]=useState('')
+    const [name, setName]=useState('');
+    const [email,setEmail]=useState('');
+    const [password, setPass]=useState('');
```
ThinkTink: *setPassword

ThinkTink (approved):
Changes look good! Just a few more small cleanup things, then it's good to merge

client/src/components/Login.js
holley2032: Sounds good!

holley2032:
@ThinkTink I incorporated those tweaks, good to merge?

ThinkTink:
> @ThinkTink I incorporated those tweaks, good to merge?

Yup, go for it @holley2032 !

PR #34, opened 2020-10-21
basic login functionality added

Mila-ashtari:
--What it does--
user is able to login to the dashboard . login functionality does not yet make server call.
In the dashboard user is able to logout and go back to sign in page.
--What to look for--
styling of the login and logout links still needs work
this is related to my previous task, but I don't know how to access the styling of the components within a material-ui component
login functionality should be integrated whit the back-end
 In the last commit user login and logout functionalities are added using context. the signup page is not yet connected to the context but works fine with singing u the user


client/debug.log
```diff
@@ -0,0 +1 @@
+[1020/145410.206:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
```
ThinkTink, line 1: Something like this can be excluded from your commits

client/src/components/Login.js
```diff
@@ -55,24 +55,24 @@ signupLink:{
...
-                    <form>
+                    <form onSubmit={props.handleLogin}>
                         <h2 className={classes.h2}>Sign in</h2>
                         <label>Your email address:</label>
-                        <TextField className={classes.textField} variant="outlined" label="email"  fullWidth required type="email"  onChange={(e)=>setEmail(e.target.value)}/>
+                        <TextField className={classes.textField} variant="outlined" label="email"  fullWidth type="email"  onChange={(e)=>setEmail(e.target.value)}/>
                         <label>Password:</label>
-                        <TextField className={classes.textField} variant="outlined" label="password" fullWidth required type="password" onChange={(e)=>setPass(e.target.value)}/>
-                        <Button className={classes.button} variant="contained" color="secondary" >Login</Button>
+                        <TextField className={classes.textField} variant="outlined" label="password" fullWidth type="password" onChange={(e)=>setPass(e.target.value)}/>
```
ThinkTink: We should keep these as required fields

client/src/components/ProfileMenu.js
```diff
@@ -43,7 +44,7 @@ const ProfileMenu = () => {
         open={Boolean(anchorEl)}
         onClose={handleClose}
       >
-        <MenuItem>Logout</MenuItem>
+        <MenuItem><Link to="/">Logout</Link></MenuItem>
```
ThinkTink: This should trigger handleLogout as well

client/src/components/ProtectedRoute.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Route } from 'react-router-dom';
+
+const ProtectedRoute = ({ component: Component, user, ...rest }) => {
+  return (
+    <Route {...rest} render={
+      props => {
+        if (user) {
+          return <Component {...rest} {...props} />
+        }
```
ThinkTink, line 12: If user is not signed in, we should redirect to the sign in page

ThinkTink (commented):
A couple small comments; I'll take another look once the user context is added!

client/src/components/ProtectedRoute.js
morgan-sam: This could be implemented with a ternary operator:
`user ? <Component  /> : <Login />`

client/debug.log
morgan-sam: If you create a `.gitignore` in the `/client/` folder and add the line:
`debug.log`
to the file, it will stop adding this to commits.

client/src/components/Login.js
```diff
@@ -101,17 +101,17 @@ function Login (){
...
                 <Box className={classes.formContainer}>
-                    <form>
+                    <form onSubmit={props.handleLogin}>
                         <h2 className={classes.h2}>Sign in</h2>
                         <label>Your email address:</label>
                         <TextField className={classes.textField} variant="outlined" label="email"  fullWidth required type="email" error={!!errors.email} helperText={errors.email} onChange={(e)=>setEmail(e.target.value)}/>
                         <label>Password:</label>
                         <TextField className={classes.textField} variant="outlined" label="password" fullWidth required type="password" error={!!errors.password} helperText={errors.password} onChange={(e)=>setPassword(e.target.value)}/>
-                        <Button className={classes.button} variant="contained" color="secondary" onClick={handleClick}>Login</Button>
+                        <Button className={classes.button} type='submit' variant="contained" color="secondary" onClick={handleClick}><Link to="/dashboard">Login</Link></Button>
```
holley2032: Would be good if the Sign Up page had this <Link to> as well!

holley2032: Assuming that the sign up is valid, of course!

holley2032 (commented):
I'm not an expert on route redirects so I can't say whether or not this is correct, but login does send the user to /dashboard (although it isn't loading on my screen, I wonder if that's a bug introduced somewhere along the way). I think the recent merge into dev from feature-auth-integration should provide the login/sign up functionalities with the back-end needed to finish up the redirecting, let me know if there's anything I can do to help!

client/src/App.js
```diff
@@ -1,26 +1,71 @@
...
       <CssBaseline>
         <MuiThemeProvider theme={theme}>
-          <BrowserRouter>
-            {/* <Route path="/" component={DashboardPage} /> */}
-            <Route path="/" component={SignUp} exact/>
-            <Route path="/login" component={Login}/>
-          </BrowserRouter>
+          <UserStore>
+            <BrowserRouter>
+              <Route path="/" exact render={()=><Login></Login>}/>
```
ThinkTink, line 20: Do we have a reason for doing render here and component below?

client/src/contexts/UserContext.js
```diff
@@ -0,0 +1,48 @@
+import React, { useState } from 'react'
+const UserContext=React.createContext({});
+
+export function UserStore(props) {
+    const [user, setUser] = useState(false)
+
+    const handleLogin = (email, password) => {
+        if (true) {
```
ThinkTink, line 8: What purpose does this if statement serve?

client/src/components/Login.js
```diff
@@ -55,67 +57,36 @@ signupLink:{
...
-    const [errors, setErrors]=useState({});
-
-    const validations = () => {
-        const errorsCopy = {...errors};
-        errorsCopy.email = email ? "" : "Please enter an email.";
-        errorsCopy.password = password ? "" : "Please enter a password.";
-        setErrors({ ...errorsCopy });
-
-        return Object.values(errorsCopy).every(field => field === "");
-    }
```
ThinkTink, line 71: We need to keep the error validation though

client/src/components/Login.js
```diff
@@ -55,67 +57,36 @@ signupLink:{
...
 
     return(
             <section className={classes.login}>
                 <Box className={classes.formContainer}>
-                    <form>
-                        <h2 className={classes.h2}>Sign in</h2>
-                        <label>Your email address:</label>
-                        <TextField className={classes.textField} variant="outlined" label="email"  fullWidth required type="email" error={!!errors.email} helperText={errors.email} onChange={(e)=>setEmail(e.target.value)}/>
-                        <label>Password:</label>
-                        <TextField className={classes.textField} variant="outlined" label="password" fullWidth required type="password" error={!!errors.password} helperText={errors.password} onChange={(e)=>setPassword(e.target.value)}/>
```
ThinkTink, line 109: and keep the error handling here too

client/src/components/Login.js
```diff
@@ -55,67 +57,36 @@ signupLink:{
...
-                        <Button className={classes.button} variant="contained" color="secondary" onClick={handleClick}>Login</Button>
-                    </form>
+                            <form onSubmit={handleSubmit}>
+                            <h2 className={classes.h2}>Sign in</h2>
```
ThinkTink, line 76: Let's use Typography here

client/src/components/ProfileMenu.js
```diff
@@ -25,11 +28,17 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-const ProfileMenu = () => {
+const ProfileMenu = (props) => {
   const [anchorEl, setAnchorEl] = useState(null);
   const classes = useStyles();
+  const value=useContext(UserContext)
```
ThinkTink, line 34: Missing semi colons here and below. Can you check that your linter is set to format on save? 
(I just had to fix mine today in VS Code, so there might have been an issue with the Prettier extension)

client/src/components/ProtectedRoute.js
ThinkTink: Don't forget this too ^

client/src/contexts/UserContext.js
```diff
@@ -0,0 +1,48 @@
...
+                        console.log(response.message);
+                    }
+                })
+                .catch((error) => {
+                    console.error('Error:', error)
+                })
+            }
+      }
+      const handleLogout = () => {
+        setUser(false);
```
ThinkTink, line 37: We'll eventually need to make an API call here in handleLogout, to clear the cookies.

We also need a handleSignup function as well.

ThinkTink (commented):
Making progress! Needs a few more features, so let's not worry about trying to get this merged before the presentation.

client/src/components/Login.js
Mila-ashtari: I put required on textField instead to require user input.

PR #35, opened 2020-10-21
Feature scraping results

morgan-sam:
**Branch dependencies:**
- Feature Amazon Scraping Service #31

**What it does:**:

Allows user to enter an Amazon item URL to the add item input form. When the add item button is clicked a popup will show. A loading spinner is shown while the request is sent to the backend and the scraper is running. Once there is a response the item or error message is shown in the popup. 

**What to look for:**

- _Most importantly_: do amazon items display correctly? I've found that various items (ebooks for example) use a different markup and so do not currently work with the scraper. Items on the daily deals use a different markup and have been accounted for. Let me know if you find any product types or listings that do no currently work with the scraper.
- The integration routing
- The UI styling/spacing
- Where to put the `getItem` function and `emptyItem` object in `AddItem.js`

**What to avoid:**

- No functionality to add items to lists
- The add new item button on the popup does nothing
- Input link only tested if it has a length greater than zero:
    - Regex testing the input URLs would be a nice feature to prevent unnecessary fetch requests


client/src/components/AddItem.js
```diff
@@ -54,9 +55,40 @@ const useStyles = makeStyles((theme) => ({
...
+};
+
 const AddItem = () => {
-  const [selectedItem, setSelectedItem] = useState("none");
+  const [inputLink, setInputLink] = useState("");
+  const [item, setItem] = useState(emptyItem);
+  const [popupOpen, setPopupOpen] = useState(false);
+  const [selectedList, setSelectedList] = useState("none");
   const classes = useStyles();
+  const openPopup = () => setPopupOpen(true);
```
ThinkTink, line 64: getItem can just live here inside the component. We can just use an empty object instead of emptyItem.

client/src/components/ItemDisplay.js
```diff
@@ -0,0 +1,82 @@
...
+  },
+  itemPrice: {
+    display: "inline",
+    color: "red",
+    fontWeight: "bold",
+  },
+}));
+
+const ItemDisplay = (props) => {
+  const { item } = props;
```
ThinkTink, line 58: Here, if we take out emptyItem above, you can define each part of item. Something like:
```
const ItemDisplay = ({ item }) => {
  const title = item.title ? item.title : ""
  const price = item.price ? item.price : ""
```

client/src/components/AddItem.js
ThinkTink, line 64: let's make openPopup() be `changePopup = () => setPopupOpen(!popupOpen)`. Then we can pass changePopup (or better name of your choosing) down to AddItemPopup. It's best practice in React not to call a setState function from outside the component where that state lives.

client/src/components/AddItem.js
```diff
@@ -54,9 +55,40 @@ const useStyles = makeStyles((theme) => ({
...
+  const addButtonClick = async () => {
+    // regex check inputLink here
```
ThinkTink: The common practice is to preface comments like this with "TODO:". Also, we may want to also add in the future to check that the link has a domain that we are able to scrape

client/src/components/AddItem.js
```diff
@@ -54,9 +55,40 @@ const useStyles = makeStyles((theme) => ({
...
+    if (inputLink.length > 0) {
+      openPopup();
+      const newItem = await getItem(inputLink);
+      setItem(newItem);
+    }
+  };
+
+  useEffect(() => setItem(emptyItem), [inputLink]);
```
ThinkTink: If someone happens to by typing in a link rather than paste it, this could get called way too many times. Would it make more sense to reset item when the popup is closed?

server/api/scraper.py
```diff
@@ -0,0 +1,49 @@
...
+        driver = webdriver.Chrome(executable_path=executable_path, options=chrome_options)    
+        driver.get(URL)
+        self.shopURL = URL
+        self.title = self.getTitle(driver) 
+        self.oldPrice = self.getOldPrice(driver)  
+        self.price = self.getPrice(driver)  
+        self.imgURL = self.getImgURL(driver)  
+        self.availability = self.getAvailability(driver)  
+        driver.quit()
+    def getTitle(self, driver):
```
ThinkTink: Python style best practice is snake_case for function names, so get_title, get_old_price, etc.

client/src/pages/AddItemPopup.js
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <Modal
+      className={classes.popup}
+      open={popupOpen}
+      onClose={handleClose}
+      aria-labelledby="add-item-popup"
+      aria-describedby="confirms-adding-item-of-url-to-selected-list"
+    >
+      <Box className={classes.paper}>
+        <h2 className={classes.paperTitle}>Add new item:</h2>
```
ThinkTink: Can we use Typography here?

ThinkTink (commented):
Left some cleanup comments on the code. Let's get some others in the group to do some testing with different links before we merge

server/api/scraper.py
```diff
@@ -0,0 +1,49 @@
...
+            return driver.find_element_by_id('productTitle').text
+        except: return None
+    def getOldPrice(self, driver):
+        try: 
+            return driver.find_element_by_class_name('priceBlockStrikePriceString').text
+        except: return None
+    def getPrice(self, driver):
+        try: 
+            return driver.find_element_by_id('priceblock_ourprice').text
+        except: return None
```
adams-liu, line 58: This is probably going to be a future PR, but the prices will be read in type float in the database rather than type string. This is good for now tho!

adams-liu (commented):
I've tested it on my local machine, and it seems to work fine! It is somewhat slow, but it's still functional which is good. There is a case where if the amazon item, has a price range (eg. $14.99-$24.99) it will show both of them as the price. Maybe we can take
the lower price or get the average?

adams-liu:
>   * Regex testing the input URLs would be a nice feature to prevent unnecessary fetch requests

Just a note for the future:
I figured a few ways to do it is "have www.amazon*", and each amazon url that is of an item has "/dp/"

server/api/scraper.py
morgan-sam: Ah good catch, I forgot to mention this in the PR. Yeah for the demo it will work as just string

Afterwards we'll have to decide on a way to store currency in the database, [but it can't be float because of floating point errors](https://stackoverflow.com/a/3730249). I've read a good way is storing it in cents as `int` and then just multiplying it up to dollars, but we can discuss this later.

client/src/components/AddItem.js
morgan-sam: Understood, however the Modal component has an `onClose` parameter rather than `onChange`.

Would it be best instead to pass a closePopup function that is:

`closePopup = () => setPopupOpen(false)`

down to AddItemPopup?

That way it's concise in its functionality, but then also the setState won't leave the parent component.

client/src/components/AddItem.js
morgan-sam: Yeah that would make more sense. The only downside would be if the link stayed the same and the popup reopened it would send the same request, but I think that's preferable to state updates on every keystroke. I'll add the reset item on close popup functionality.

client/src/components/AddItem.js
morgan-sam: Ok combining your other suggestion below I've changed AddItem to instead pass down:
```
const closePopup = () => {
    setPopupOpen(false);
    setItem({});
};
```

to AddItemPopup.

ThinkTink: Technically we can pass an onClose any function we want, but given that we're also clearing the setItem on close this is cleaner!

ThinkTink (approved):

morgan-sam:
> I've tested it on my local machine, and it seems to work fine! It is somewhat slow, but it's still functional which is good. 

It'll be a bit slow as selenium has to startup for each scrape, but once we have the browser persist it will be much faster.

>There is a case where if the amazon item, has a price range (eg. $14.99-$24.99) it will show both of them as the price. Maybe we can take the lower price or get the average?

Oh yeah, that needs to be fixed. I'd opt for the lower price but it can be inaccurate. Sometimes vendors game it by adding irrelevant low cost items to listings so they can appear in _sort price by low to high_; so average may be better.

> Just a note for the future:
> I figured a few ways to do it is "have www.amazon*", and each amazon url that is of an item has "/dp/"

Yeah that would be the way to do it. In terms of storing the link on the database we could strip the URL of the extra unneeded info, so something like:

https://www.amazon.com/Microphone-Streaming-Computer-Conferencing-Facetime/dp/B087NN41JH?ref_=Oct_DLandingS_D_4522958b_NA

Would only be stored as:

https://www.amazon.com/dp/B087NN41JH

aeciorc (commented):
Looks great! Two things I noticed while testing:
- Passing `chrome_type=ChromeType.CHROME` to the ChromeDriverManager caused an issue for me. It might be specific to MacOS. When passing no args, it worked fine (it defaults to google-chrome)
-Images were not visible in the results (src=unknown)

aeciorc (approved):

PR #43, opened 2020-10-23
Fixed login & signup routing

morgan-sam:
**What it does:**

- Protected route with `value.user` returns Component and without returns Login
- Signup on success shows alert and redirects to "/"
- handleLogin function returns success/fail boolean which determines if to redirect to "/dashboard"

**What to avoid:**

- No features added, only changes redirects for login/signup

**Note:**

- Changes are very minimal; Prettier formatting makes changes look much larger

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -1,14 +1,14 @@
-import React, { Fragment, useState } from "react";
+import React, { Fragment } from "react";
 import { MuiThemeProvider } from "@material-ui/core";
 import CssBaseline from "@material-ui/core/CssBaseline";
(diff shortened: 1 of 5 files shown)


holley2032:
It works well on my end, redirects are good, though is it possible to redirect directly from a successful signup to "/dashboard" without introducing a bug? I know it might be difficult to do before the demo, so no worries if not, this pull request works well as is!

holley2032 (approved):

adams-liu (approved):
I got this working on my setup as well!

PR #44, opened 2020-10-24
Feature frontend products lists

holley2032:
--What it does--
Provides the basic front-end framework for adding a product via list details, and adding a list.

--What it avoids--
It does not integrate anything with the back-end.
The styling leaves much to be desired. I'm not particularly experienced with the front-end, and I'm happy to try to fix what's wrong with it, but I think it's better to get this pushed forward now and focus on other tasks that are more pressing and might need to be completed to free up other tasks.

--What to look for--
If there's any problems with the visuals that can be fixed relatively easily.


client/src/components/AddList.js
```diff
@@ -0,0 +1,231 @@
...
+  },
+  popup: {
+    display: "flex",
+    alignItems: "center",
+    justifyContent: "center",
+  },
+  paper: {
+    position: "absolute",
+    display: "flex",
+    //display:'block',
```
morgan-sam: Make sure to remove all commented out styles before committing

client/src/components/AddList.js
```diff
@@ -0,0 +1,231 @@
...
+  centerText:{
+    textAlign:"center"
+  },
+  imageFieldContainer:{
+    display:"flex",
+    backgroundColor:"white",
+    width:"35%",
+    height:"0",
+    paddingBottom:"35%"
+  },
```
morgan-sam: To improve the appearance of the imageFieldContainer I would:
- Change the 2 inner text lines to 2 separate ` <Typography />` components (explained later in review)
- Remove the paddingBottom
- Give it a fixed height (13rem x 13rem looked good to me)
- Style it as: `display: "flex", alignItems: "center", justifyContent: "center", flexDirection: "column",`
- Add a `border: 1px dashed grey` and give it a border radius of 1rem

This will center the text on 2 separate lines for a neater appearance. Also the dotted line should give the box a "cut-out" look that works as a visual cue to help suggest to the user that they should drag and place an image there.

client/src/components/AddList.js
```diff
@@ -0,0 +1,231 @@
...
+              aria-label="close"
+              className={classes.closeButton}
+              onClick={handleClose}
+              size="small"
+            >
+              <CloseIcon fontSize="small" />
+            </IconButton>
+          </Box>
+          <Box className={classes.titleContainer}>
+            <Typography><h2 className={classes.titleText}>Create new list</h2></Typography>
```
morgan-sam: When using the `<Typography />` component default HTML text tags as children are unnecessary.

Instead for instance you can do:
`<Typography variant="h2">Create new list</Typography>`

Or just attach the class straight to the Component:
`<Typography className={classes.titleText}>Create new list</Typography>`

client/src/components/AddList.js
```diff
@@ -0,0 +1,231 @@
...
+            <TextField
+              className={classes.textField}
+              variant="outlined"
+              placeholder="Enter name" // Placeholder needs to be centered.
+              fullWidth
+              type="text"
+              error={!!errors.title}
+              helperText={errors.title}
+              onChange={(e)=>setTitle(e.target.value)}/>
+          </Box>
```
morgan-sam: I would place the `<Typography />` title and the `<TextField />` in the same `<Box />` container as they should be grouped together.

Then I would add a small amount of margin to the box, to space it out from the rest of the popup elements.

client/src/components/AddList.js
```diff
@@ -0,0 +1,231 @@
...
+              accept="image/*"
+              className={classes.input}
+              id="file-input"
+              multiple
+              type="file"
+              />
+            <label htmlFor="file-input">
+              select a file
+            </label></u></strong></Typography>
+          </Box>
```
morgan-sam: Same as above, both `imageFieldContainer` and the `<Typography />` title should be surrounded by a single `<Box />` and given a small amount of margin (imageFieldContainer itself should still be a `<Box />`).

client/src/components/ListsDisplay.js
```diff
@@ -53,20 +53,22 @@ const demoListsArray = [
...
+
   return (
     <Box className={classes.shoppingContainer}>
       <Typography variant="h5" className={classes.listsTitle}>
         My Shopping Lists:
       </Typography>
       <Box className={classes.myShoppingLists}>
         {demoListsArray.map((list, i) => (
-          <ListCard key={i} list={list} />
+          <ListCard key={i} list={list} demoListsArray={demoListsArray}/>
```
morgan-sam: You can shorten this line to:
`<ListCard key={i} {...{ list, demoListsArray }} />`

client/src/pages/ShoppingLists.js
```diff
@@ -1,13 +1,16 @@
...
+import AddList from "components/AddList"
 
 const ShoppingLists = () => {
+    const [addListOpen, setAddListOpen] = useState(false)
   return (
     <Box>
       <AddItem />
-      <ListsDisplay />
+      <ListsDisplay addListOpen = {addListOpen} setAddListOpen = {setAddListOpen} />
+      <AddList addListOpen = {addListOpen} setAddListOpen = {setAddListOpen}/>
```
morgan-sam: Can both also be shortened to:
`<Component {...{ addListOpen, setAddListOpen }} />`

client/src/components/AddProduct.js
```diff
@@ -0,0 +1,249 @@
...
+            <Typography className={classes.itemText}>Paste link to item:</Typography>
+          </Box>
+          <Box className={classes.textFieldContainer}>
+            <TextField
+              className={classes.textField}
+              variant="outlined"
+              placeholder="Paste your link here" // Placeholder needs to be centered.
+              fullWidth
+              type="text"
+              error={!!errors.title}
```
morgan-sam: Assuming `errors.title` is a boolean; `!!errors.title === errors.title`

Is this a mistake?

morgan-sam: Ahh I think I understand what you want.

You are looking for: `error={Boolean(errors.title)}`

This is a lot clearer in terms of intention, instead of a double negative.

client/src/components/AddProduct.js
```diff
@@ -0,0 +1,249 @@
...
+    justifyContent: "center",
+    backgroundColor: "white",
+    margin: "3rem",
+    overflow: "hidden",
+    height: "4rem",
+    padding: "0 0.5rem",
+    boxShadow: "0 0 0.5rem 1px #eee",
+    width: "80%"
+  }
+}));
```
morgan-sam: Styles way over 100 lines are best to put in a separate file and import in.

client/src/components/ListCard.js
```diff
@@ -42,12 +43,17 @@ const useStyles = makeStyles((theme) => ({
...
+  const {list, demoListsArray} = props;
   const { listTitle, itemCount, img } = props.list;
   const classes = useStyles();
   const [listOpen, setListOpen] = useState(false);
   const changeListOpen = () => {
     setListOpen(!listOpen);
   };
+  const [addProductOpen, setAddProductOpen] = useState(false);
+  const changeAddProductOpen = () => {
+    setAddProductOpen(!addProductOpen);
```
morgan-sam: With state manipulation that is dependent on the previous state (i.e. toggling a boolean, like this case) it's best to use the following syntax:

`setAddProductOpen((previous) => !previous);`   (where "previous" can be named anything)

This ensures the previous state value is used, as react setState is asynchronous so it is not usually guaranteed that the previous value will be updated in time ([explained here](https://stackoverflow.com/a/36087156)).

morgan-sam (changes requested):
Make sure you are using Prettier with React files otherwise our commits will clash.

It's almost there, let's improve the visuals before merging. Once the `imageFieldContainer` is fixed it should basically be ready to merge.

client/src/components/AddList.js
morgan-sam: For the text of `<imageFieldContainer />`, I would structure it the following way:

```
<Box className={classes.imageFieldContainer}>
  <Typography>Drop an image here or </Typography>
  <Typography>
    <input ... />
    <label ... >select a file</label>
  </Typography>
</Box>
```

That way when you style `imageFieldContainer` to use `flexDirection: "column"`, the text will both sit on separate lines.

Also add a class to `<label>` and use css to make it bold and underlined instead of using HTML tags. The `<u>` and `<strong>` tags are usually considered obsolete.

client/src/components/AddProduct.js
```diff
@@ -0,0 +1,249 @@
...
+
+const AddList = (props) => {
```
ThinkTink: Be sure to update function name here and at export (line 249) to AddProduct

client/src/components/ListCard.js
```diff
@@ -42,12 +43,17 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 const ListCard = (props) => {
+  const {list, demoListsArray} = props;
   const { listTitle, itemCount, img } = props.list;
```
ThinkTink: Can we combine these two lines?

client/src/components/AddProduct.js
```diff
@@ -0,0 +1,249 @@
...
+          <Box className={classes.textFieldContainer}>
+            <TextField
+              className={classes.textField}
+              variant="outlined"
+              placeholder="Paste your link here" // Placeholder needs to be centered.
+              fullWidth
+              type="text"
+              error={!!errors.title}
+              helperText={errors.title}
+              onChange={(e)=>setTitle(e.target.value)}/>
```
ThinkTink: This field is for the link, right? Maybe a variable name of 'url' would be more clear than 'title'?

client/src/pages/ShoppingLists.js
```diff
@@ -1,13 +1,16 @@
-import React from "react";
+import React, { useState } from "react";
 import { Box } from "@material-ui/core";
 import AddItem from "components/AddItem";
 import ListsDisplay from "components/ListsDisplay";
+import AddList from "components/AddList"
 
 const ShoppingLists = () => {
+    const [addListOpen, setAddListOpen] = useState(false)
```
ThinkTink: Like we've done elsewhere, let's not pass setAddListOpen directly to the child components; instead add a function here that calls setAddListOpen and pass that function down

client/src/components/AddList.js
```diff
@@ -0,0 +1,231 @@
...
+  input: {
+    display:'none'
+  },
+}));
+
+const AddList = (props) => {
+    const {addListOpen, setAddListOpen} = props;
+    const classes = useStyles()
+    const [errors, setErrors]=useState({});
+    const [title, setTitle]=useState('');
```
ThinkTink: Let's be consistent with spacing around the =

client/src/components/AddProduct.js
```diff
@@ -0,0 +1,249 @@
...
+    const [title, setTitle]=useState('');
+    const [list, setList]=useState('');
+    const [errors, setErrors]=useState({});
+
+    const handleChange = (event) => {
+      setList(event.target.value);
+    };
+
+    const handleClose = () => {
+        changeAddProductOpen();
```
ThinkTink: Looks like indents aren't consistent. Can you check your prettier settings?

holley2032: It wasn't on when I initially made this, thanks!

client/src/components/ListCard.js
holley2032: I'm not sure, I tried a few different things and it seemed like I couldn't get it to work unless I used two lines like this. I'd definitely prefer it to be one line if possible!

holley2032:
@morgan-sam @ThinkTink Thank y'all so much for your comments, they were really helpful! I've updated the code accordingly, let me know if it looks good to merge.

client/src/components/ListCard.js
ThinkTink: I think it'd be `const { { listTitle, itemCount, img }, demoListsArray} = props`, or maybe `const { { ...listTitle, itemCount, img }, demoListsArray} = props`.... but if we can't get it to work, definitely not a problem to do it in two lines

client/src/components/AddList.js
```diff
@@ -0,0 +1,125 @@
...
+            <Typography>
+              <input
+                accept="image/*"
+                className={classes.input}
+                id="file-input"
+                multiple
+                type="file"
+              />
+              <label htmlFor="file-input" className={classes.imageFieldText}>
+                select a file
```
ThinkTink, line 109:
```suggestion
                Select a file
```

client/src/components/AddProduct.js
```diff
@@ -0,0 +1,149 @@
...
+    e.preventDefault();
+    if (validations()) {
+      fetch("route", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+          url: url,
+          //'cover': cover
```
ThinkTink:
```suggestion

```

client/src/components/AddList.js
```diff
@@ -0,0 +1,125 @@
...
+        <Box className={classes.bodyContainer}>
+          <Typography className={classes.itemText}>Add a title</Typography>
+          <Box className={classes.textFieldContainer}>
+            <TextField
+              className={classes.textField}
+              variant="outlined"
+              placeholder="Enter name" // Placeholder needs to be centered.
+              fullWidth
+              type="text"
+              error={!!errors.title}
```
ThinkTink: Let's keep this consistent throughout (`Boolean(errors.title)`)

client/src/components/ListCard.js
holley2032: I tried both of those suggestions, and neither worked, so I'll keep it as is for now.

client/src/components/AddList.js
holley2032: Because this is part of a sentence with the Typography above, it will read "Drop an image here or select a file" over two lines, so I think it makes sense to keep this lowercase.

holley2032:
Fixed those errors mentioned by @ThinkTink, I may have missed some other things but I think it should be good now!

ThinkTink (approved):

PR #51, opened 2020-10-26
Feature price history

morgan-sam:
**What it does**

The price routing of the app, updated Price model, the readme of the routes, and the scraper now takes the smallest URL it can.

**Routes:**
  - All products:
     - GET route for price history of all products
   - Single product:
       - GET route for getting price history of single product
       - DELETE route for deleting price history of single product
       - POST route for posting new price of a single product

**What to look for**

- Is the routing correct? Would we need any other routes?
- Prices model: now using product_id (setup to be a ForeignKey)
- The [readme docs for the price routes](https://github.com/hatchways/team-apple-cider/blob/aa5c29694a6bb3154ec95d3e3763e8154d3f2070/README.md#Prices): Are they clear, do they make sense, and do the examples work?
- Are there any valid long URLs that do not standardize down to a smaller valid URL? 

**What not to look for**

- Price routes are not yet called from the scraper. I need to implement a way of identifying the corresponding product from the scraper. This should be more feasible now as all URLs are standardized down to a smaller format.


server/api/price_history_handler.py
```diff
@@ -0,0 +1,44 @@
...
+@price_history_handler.route('/price_history', methods = ['GET', 'POST'])
+def price_history():
+    if request.method == 'GET':
+        price_history = Prices.query.all()
+        price_history_ls = []
+        for item in price_history:
+            price_history_ls.append(
+                {
+                    "id": item.id,
+                    "url_id": item.url_id,
```
adams-liu: what is `url_id` is that `product_id` , you don't technically need the `url_id` unless it's that case in which the product will be changing urls, in which case the user would need to input the new url

server/api/price_history_handler.py
```diff
@@ -0,0 +1,44 @@
...
+                    "price": item.price,
+                    "scrape_date": item.scrape_date
+                })
+        return jsonify({'prices': price_history_ls}), 200
+    if request.method == 'POST':
+        URL = json.loads(request.get_data())['url']
+        try:
+            item = ScrapeAmazon(URL)
```
adams-liu: instead of scraping amazon instantly, it might be smarter to check if an entry already exists in the `price_history` table before scraping it. Obviously, we would need to check when it was last scraped.

server/models/prices.py
```diff
@@ -0,0 +1,16 @@
+from database import db
+import datetime
+
+class Prices(db.Model):
+
+    __tablename__ = 'price_history'
+
+    id = db.Column(db.Integer, primary_key=True, unique=True, autoincrement=True)
```
adams-liu: we don't need a primary key on this table, it doesn't really do much for us

server/models/prices.py
```diff
@@ -0,0 +1,16 @@
...
+    url_id = db.Column(db.String, nullable=False)
```
adams-liu: change to a `product_id`, and make sure it's a foreign key

server/sql/price_history.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE price_history (
+  id SERIAL PRIMARY KEY,
+  url_id VARCHAR(255) NOT NULL,
```
adams-liu: change `url_id`

adams-liu (changes requested):

server/api/price_history_handler.py
adams-liu: sorry, I just realized, you tried to do a regex, to get the url id based on the amazon link

adams-liu: this still doesn't make sense because we still need a way to link between the product table and the `prices_history` table, that's why we need `product_id`. I think the scrape URL code makes sense maybe in the scraper code to check if it's an amazon link?

server/models/prices.py
```diff
@@ -0,0 +1,16 @@
...
+    price = db.Column(db.Integer, nullable=False)
```
adams-liu: should be type `numeric`

server/api/scraper.py
```diff
@@ -3,6 +3,10 @@
 from selenium.webdriver.chrome.options import Options
 from webdriver_manager.chrome import ChromeDriverManager
 
+def string_to_int_price(price_string):
```
adams-liu, line 6: `string_to_decimal_price`

server/api/prices_handler.py
```diff
@@ -0,0 +1,44 @@
...
+from flask import jsonify, request, Blueprint
+from models.price import Price
+from database import db
+prices_handler = Blueprint('prices_handler', __name__)
+from .scraper import ScrapeAmazon
+
+
+def get_url_id(URL):
+    x = re.search(r"amazon.com\/.*dp\/([^/]+)", URL) 
+    return x.group(1)
```
ThinkTink: Let's use a more descriptive variable name than 'x'

server/api/prices_handler.py
```diff
@@ -0,0 +1,44 @@
...
+    
+
+@prices_handler.route('/prices', methods = ['GET', 'POST'])
+def prices():
+    if request.method == 'GET':
+        prices = Price.query.all()
+        prices_ls = []
```
ThinkTink: prices_list would be better

server/api/prices_handler.py
```diff
@@ -0,0 +1,44 @@
...
+        for item in prices:
+            prices_ls.append(
+                {
+                    "id": item.id,
+                    "url_id": item.url_id,
+                    "price": item.price,
+                    "scrape_date": item.scrape_date
+                })
+        return jsonify({'prices': prices_ls}), 200
```
ThinkTink: Do we want all prices in the entire database? Or maybe just all prices given an item id?

server/api/scrape_handler.py
```diff
@@ -6,7 +6,7 @@
 @scrape_handler.route('/scrape', methods = ['POST'])
 def scrape():
     if request.method == 'POST':
-        URL = json.loads(request.get_data())['URL']
+        URL = json.loads(request.get_data())['url']
```
ThinkTink: Every time we scrape anything we'd want to add its price to the price table. Currently, we call ScrapeAmazon at /scrape and at /prices, but price only gets saved on the later.

Perhaps we have this route call POST to /prices, and remove the separate call to ScrapeAmazon in that route. Then when we have it set up to scrape our urls every X# hours, we could just call /scrape. We don't need it set up like that necessarily, but let's be sure we're always saving needed data without doing redundant work

server/models/prices.py
morgan-sam: SQLAlchemy doesn't seem to work with tables that don't contain primary keys.

Would it be best instead differentiate by having a `primary_key` and a `product_key`?

The primary key would refer to the individual price entry and the product key the single product.

ThinkTink: yup, every table in a SQL-like database needs a primary key. It's the way to uniquely identify each row. Then product_key would be our foreign key, connecting this table to the product table.

client/src/components/ItemDisplay.js
```diff
@@ -55,6 +55,12 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+const centsToDollarsDisplay = (inputCents) => {
+  const dollars = String(Math.floor(inputCents / 100));
+  const cents = String(inputCents % 100);
+  return `$${dollars}.${cents.length > 1 ? cents : cents + "0"}`;
```
ThinkTink: Shouldn't it be `: "0" + cents`?

server/api/prices_handler.py
```diff
@@ -0,0 +1,53 @@
...
+        return jsonify({'response': "Product '{}' prices were successfully deleted from the database".format(product_id)}), 200
+    
+
+@prices_handler.route('/prices', methods = ['GET'])
+def allPriceRequests():
+    if request.method == 'GET':
+        prices = Price.query.all()
+        prices_list = []
+        for item in prices:
+            prices_list.append(
```
ThinkTink: Can we use the item.serialize method added to the price model?

client/src/components/ItemDisplay.js
morgan-sam: ahh good catch

PR #53, opened 2020-10-28
Bugfix auth integration

holley2032:
--What it does--
Fixes bug where refreshing dashboard logs user out.
Causes logout button to delete cookie, using a backend logout API.
Adjusts styling on signup page to accommodate tooltip messages.
Adds a tooltip message to warn a user that their input is invalid on confirm password.
Uses HTML alerts on other front-end based invalid inputs
Adds snackbars on signup page when the back-end throws an error.
Adds snackbars on the login page.
Introduces minor bug where a logged out user can briefly see the dashboard when "/dashboard" is the route.
Introduces minor bug where a user may be redirected to the login page briefly before going to the dashboard when signing up.
Implements the framework for a landing page and associated dialogs.
Blacklists authentication tokens when a user logs out and adds the associated database model.
Checks if an authentication token is blacklisted when a user logs out.

--What it avoids--
It doesn't style the landing page much at all. I was going to add a close button to the dialogs opened by the sign up and login buttons, but I will be busy the rest of today and I felt like that would be a very minor thing to hold up this pull request for, especially since this pull request already covers a fair bit of ground, and the goal was more to provide a skeleton for the landing page rather than a full-fledged landing page.

--What to look for--
If the styling works, especially on the warnings now given when users input info that is wrong or invalid.
If there are any bugs not already noted.
If there is a way to fix the minor bugs noted above.

### DIFF
--- a/client/src/components/Login.js
+++ b/client/src/components/Login.js
@@ -1,13 +1,9 @@
 import React, { useState, useContext } from "react";
 import { Link, withRouter } from "react-router-dom";
 import { makeStyles } from "@material-ui/core/styles";
-import { Button, TextField, Box, Snackbar } from "@material-ui/core";
-import MuiAlert from "@material-ui/lab/alert";
+import { Button, TextField, Box } from "@material-ui/core";
 import UserContext from "../contexts/UserContext";
-
-function Alert(props) {
-  return <MuiAlert elevation={6} variant="filled" {...props} />;
-}
+import WarningSnackbar from "./WarningSnackbar";
 
 const useStyles = makeStyles((theme) => ({
   login: {
@@ -68,11 +64,9 @@ function Login(props) {
   const [openSnack, setOpenSnack] = useState(false);
   const [snackText, setSnackText] = useState("");
   const value = useContext(UserContext);
-  const vertical = "top";
-  const horizontal = "center";
 
-  const handleSnack = (props) => {
-    setSnackText(props);
+  const handleSnack = (message) => {
+    setSnackText(message);
     setOpenSnack(true);
   };
 
@@ -99,17 +93,13 @@ function Login(props) {
       .then((response) => response.json())
       .then(function (response) {
         if (response.status === "success") {
-          console.log("Success:", email);
           const loginSuccess = value.handleLogin(email, password);
           if (loginSuccess) props.history.push("/dashboard");
         } else {
           handleSnack(response.message);
-          console.log(response.message);
         }
       })
-      .catch((error) => {
-        console.error("Error:", error);
-      });
+      .catch(() => {});
   };
 
   return (
@@ -152,16 +142,11 @@ function Login(props) {
             Create an account
           </Link>
         </Box>
-        <Snackbar
-          open={openSnack}
-          autoHideDuration={6000}
-          onClose={handleCloseSnack}
-          anchorOrigin={{ vertical, horizontal }}
-        >
-          <Alert onClose={handleCloseSnack} severity="error">
-            {snackText}
-          </Alert>
-        </Snackbar>
+        <WarningSnackbar
+          openSnack={openSnack}
+          handleCloseSnack={handleCloseSnack}
+          snackText={snackText}
+        />
       </Box>
     </section>
   );
--- a/client/src/components/Login.js
+++ b/client/src/components/Login.js
@@ -1,9 +1,13 @@
 import React, { useState, useContext } from "react";
 import { Link, withRouter } from "react-router-dom";
 import { makeStyles } from "@material-ui/core/styles";
-import { Button, TextField, Box } from "@material-ui/core";
-
+import { Button, TextField, Box, Snackbar } from "@material-ui/core";
+import MuiAlert from "@material-ui/lab/alert";
 import UserContext from "../contexts/UserContext";
+
+function Alert(props) {
+  return <MuiAlert elevation={6} variant="filled" {...props} />;
+}
 
 const useStyles = makeStyles((theme) => ({
   login: {
@@ -61,18 +65,57 @@ function Login(props) {
   const classes = useStyles();
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
+  const [openSnack, setOpenSnack] = useState(false);
+  const [snackText, setSnackText] = useState("");
   const value = useContext(UserContext);
+  const vertical = "top";
+  const horizontal = "center";
 
-  const handleSubmit = async (e) => {
+  const handleSnack = (props) => {
+    setSnackText(props);
+    setOpenSnack(true);
+  };
+
+  const handleCloseSnack = (event, reason) => {
+    if (reason === "clickaway") {
+      return;
+    }
+
+    setOpenSnack(false);
+  };
+
+  const handleLogin = async (e) => {
     e.preventDefault();
-    const loginSuccess = value.handleLogin(email, password);
-    if (loginSuccess) props.history.push("/dashboard");
+    fetch("/auth/login", {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({
+        email: email,
+        password: password,
+      }),
+    })
+      .then((response) => response.json())
+      .then(function (response) {
+        if (response.status === "success") {
+          console.log("Success:", email);
+          const loginSuccess = value.handleLogin(email, password);
+          if (loginSuccess) props.history.push("/dashboard");
+        } else {
+          handleSnack(response.message);
+          console.log(response.message);
+        }
+      })
+      .catch((error) => {
+        console.error("Error:", error);
+      });
   };
 
   return (
     <section className={classes.login}>
       <Box className={classes.formContainer}>
-        <form onSubmit={handleSubmit}>
+        <form onSubmit={handleLogin}>
           <h2 className={classes.h2}>Sign in</h2>
           <label>Your email address:</label>
           <TextField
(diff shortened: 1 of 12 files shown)


client/src/components/SignUp.js
```diff
@@ -132,17 +140,25 @@ function SignUp(props) {
...
-            error={!!errors.name}
-            helperText={errors.name}
-            onChange={(e) => setName(e.target.value)}
-          />
+          <Tooltip
+            open={openNameTooltip}
+            title={errors.name}
+            arrow
+            disableHoverListener
+            disableTouchListener
```
Mila-ashtari: I used required attribute to make sure the inputs are filled by the user

client/src/components/SignUp.js
```diff
@@ -181,9 +197,10 @@ function SignUp(props) {
           />
           <Button
             className={classes.button}
+            onClick={handleClick}
+            //type="submit"
```
Mila-ashtari: I added onSubmit on the form element instead of onClick on the button

client/src/components/SignUp.js
Mila-ashtari: Please check the last commit on this PR to see my changes

holley2032: Looks good, thank you!

client/src/components/SignUp.js
holley2032: Looks good!

client/src/components/Login.js
```diff
@@ -61,18 +65,57 @@ function Login(props) {
...
+  const handleSnack = (props) => {
+    setSnackText(props);
+    setOpenSnack(true);
+  };
+
+  const handleCloseSnack = (event, reason) => {
+    if (reason === "clickaway") {
+      return;
+    }
+
```
adams-liu, line 77: weird extra space

client/src/components/Login.js
```diff
@@ -61,18 +65,57 @@ function Login(props) {
...
+      .then((response) => response.json())
+      .then(function (response) {
+        if (response.status === "success") {
+          console.log("Success:", email);
+          const loginSuccess = value.handleLogin(email, password);
+          if (loginSuccess) props.history.push("/dashboard");
+        } else {
+          handleSnack(response.message);
+          console.log(response.message);
+        }
```
Mila-ashtari: It seems that you have moved handle login from context to login component which makes context useless, any reason why the handle login have been moved to login component?

holley2032: I couldn't figure out how to get the response.message to handleSnack when using context. If there is a way to do that, I'm all for it.

ThinkTink: Rather than return a boolean from handleLogin in context, we could return the response object, then check its message here

client/src/components/Login.js
```diff
@@ -61,18 +65,57 @@ function Login(props) {
   const classes = useStyles();
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
+  const [openSnack, setOpenSnack] = useState(false);
+  const [snackText, setSnackText] = useState("");
   const value = useContext(UserContext);
+  const vertical = "top";
+  const horizontal = "center";
+
+  const handleSnack = (props) => {
```
ThinkTink: Since this is just within a single component, let's use 'message' or 'text' or something instead of props

client/src/components/Login.js
```diff
@@ -61,18 +65,57 @@ function Login(props) {
...
+    })
+      .then((response) => response.json())
+      .then(function (response) {
+        if (response.status === "success") {
+          console.log("Success:", email);
+          const loginSuccess = value.handleLogin(email, password);
+          if (loginSuccess) props.history.push("/dashboard");
+        } else {
+          handleSnack(response.message);
+          console.log(response.message);
```
ThinkTink: now that we have the snackbar working let's remove all the console.logs

client/src/components/SignUp.js
```diff
@@ -105,9 +167,9 @@ function SignUp(props) {
   };
 
   return (
-    <section className={classes.login}>
+    <section className={classes.signup}>
       <Box className={classes.formContainer}>
-        <form>
+        <form form onSubmit={handleSignup}>
```
ThinkTink:
```suggestion
        <form onSubmit={handleSignup}>
```

client/src/components/SignUp.js
```diff
@@ -141,37 +202,58 @@ function SignUp(props) {
...
+        <Snackbar
+          open={openSnack}
+          autoHideDuration={6000}
+          onClose={handleCloseSnack}
+          anchorOrigin={{ vertical, horizontal }}
+        >
+          <Alert onClose={handleCloseSnack} severity="error">
+            {snackText}
+          </Alert>
+        </Snackbar>
```
ThinkTink: Since this is used in two components, we could extract it as its own component.

client/src/components/SignUp.js
```diff
@@ -141,37 +202,58 @@ function SignUp(props) {
...
+              required
+              type="password"
+              error={!!errors.confirm}
+              // helperText={errors.confirm}
+              onChange={(e) => setConfirm(e.target.value)}
+            />
+          </ErrorTooltip>
           <Button
             className={classes.button}
+            // onClick={handleClick}
```
ThinkTink: Be sure to remove this and all the other commented out code above before merging

client/src/contexts/UserContext.js
```diff
@@ -34,7 +77,22 @@ export function UserStore(props) {
...
+    })
+      .then((response) => response.json())
+      .then((response) => {
+        if (response.status === "success") {
+          console.log("Logout successful");
+          setUser(false);
+        } else {
+          setUser(false);
+        }
+      })
```
ThinkTink:
```suggestion
setUser(false);
      })
```

server/api/auth_handler.py
```diff
@@ -96,6 +118,22 @@ def post(self):
             ):
                 auth_token = user.encode_auth_token(user.id)
                 if auth_token:
+                    print("hello")
```
ThinkTink:
```suggestion

```

server/api/auth_handler.py
```diff
@@ -96,6 +118,22 @@ def post(self):
...
+                    try:
+                        blacklist_check = BlacklistToken.query.filter_by(
+                            token=str(auth_token)
+                        ).first()
+                    except Exception as e:
+                        print(e)
```
ThinkTink:
```suggestion

```

server/api/auth_handler.py
```diff
@@ -96,6 +118,22 @@ def post(self):
...
+                        return "Error"
+                    print("hello")
```
ThinkTink:
```suggestion

```

client/src/components/Login.js
holley2032: I tried doing that before, the problem is that the object returned is a promise, and that promise doesn't work well with if/else statements for whatever reason. For instance, I tried expanding on the original if statement (which did work with the promise) by adding an else that used handleSnack(loginSuccess.message);, but it didn't work. I'm definitely open to using the UserContext, but I spent a fair bit of time trying to make it work with that and I couldn't figure it out. If someone has more experience handling promises, that'd be helpful!

holley2032:
@ThinkTink @Mila-ashtari I added in Carmen's suggestions, though I still couldn't figure out how to keep the UserContext involved while also getting the message to the Snackbar. I'm not super skillful when it comes to passing info between components, and I'm almost certain there's a way to fix it, I'm just not sure what that way is.

client/src/components/Login.js
ThinkTink: hmm. Yeah, promises can be tricky. Were you approaching it like this?
```
const handleLogin = async (e) => {
  const response = await value.handleLogin(email, password);
  if (response.status === "success") //do stuff
  else handleSnack(response.message);
```
if that doesn't work, why don't we make the API call here and then call handleLogin from Context just to set the user object to true - that's the important part of using context, not the actual location of the API calls

holley2032: That worked, thank you so much! I think I didn't put await in initially, and that was causing the problems.

holley2032: I'm going to try to put in something similar in SignUp.js so the code isn't repeated between UserContext and SignUp.

holley2032:
@ThinkTink @Mila-ashtari I think it should be basically good to go now, I tested out using the UserContext with the snackbars and they seem to work, thank you so much @ThinkTink for helping me understand JavaScript promises better!

client/src/components/SignUp.js
```diff
@@ -118,7 +144,6 @@ function SignUp(props) {
             required
             type="text"
             error={!!errors.name}
-            helperText={errors.name}
```
ThinkTink, line 121: If we're taking out this type of validation, be sure to remove line 120 as well as all of the code from validations() related to these fields

client/src/components/SignUp.js
```diff
@@ -141,37 +164,49 @@ function SignUp(props) {
...
+            disableFocusListener
+          >
+            <TextField
+              className={classes.textField}
+              variant="outlined"
+              label="confirm password"
+              fullWidth
+              required
+              type="password"
+              error={!!errors.confirm}
```
ThinkTink: `Boolean(errors.confirm)`

client/src/components/WarningSnackbar.js
```diff
@@ -0,0 +1,27 @@
...
+      anchorOrigin={{ vertical, horizontal }}
+    >
+      <Alert onClose={handleCloseSnack} severity="error">
+        {snackText}
+      </Alert>
+    </Snackbar>
+  );
+}
+
+export default WarningSnackbar;
```
ThinkTink, line 27: :+1:

ThinkTink (approved):
Just a couple last cleanup things and it's good to go!

client/src/components/SignUp.js
holley2032: I edited it as you recommended. I could have removed the validations() step altogether (leaving only a value for errorConfirm), but I think I'd prefer to leave them in in case we want to add back in more custom errors or make the styling consistent, which would use validations().

PR #54, opened 2020-10-28
friends page content created

Mila-ashtari:
--What it does--
friends page have followers and following tabs with their content. people can be followed or unfollowed.

--What it avoids--
friends page is still a tab , the respective route to be created .
the page is not integrated with back-end yet.


client/src/components/Followers.js
```diff
@@ -0,0 +1,56 @@
...
+                     <Button
+                        className={classes.unFollowButton}
+                        variant="contained"
+                        onClick={()=>props.handleFollow(person)}
+                        >
+                        follow
+                    </Button>
+                </ListItem>
+                <Divider />
+              </Fragment>
```
morgan-sam: As the functionality between this section and the following fragment section is very similar, it could encapsulated in a single component called something like `<ProfileList />`. The only difference is the button, which could be changed to toggle between unfollow/follow so they would both be identical.

client/src/components/Following.js
```diff
@@ -0,0 +1,55 @@
...
+                     <ListItemText>
+                         {person.name}
+                     </ListItemText>
+                     <Button
+                        className={classes.unFollowButton}
+                        variant="contained"
+                        onClick={()=>props.handleUnfollow(person)}
+                        >
+                        Unfollow
+                    </Button>
```
morgan-sam: If the button was changed to toggle it would look like this:
```
<Button
	className={classes.toggleFollowButton}
	variant="contained"
	onClick={()=>props.toggleFollow(person)}
>
	{followingsList.includes(person) ? 'Unfollow' : 'Follow'}
</Button>
```

client/src/components/Following.js
```diff
@@ -0,0 +1,55 @@
...
+        fontSize:'0.6rem'
+        
+    }
+    
+  }));
+const Followers =(props)=>{
+    const classes = useStyles();
+    return(
+        <List aria-label='following list' className={classes.list}>
+            {props.followingsList.map((person=>{
```
morgan-sam: If you deconstruct props at the top of the component it makes them easier to use later:

`const { followingsList } = props;`

client/src/components/ProtectedRoute.js
```diff
@@ -10,7 +10,7 @@ const ProtectedRoute = ({ component: Component, ...rest }) => {
     <Route
       {...rest}
       render={(props) =>
-        value.user ? <Component {...rest} {...props} /> : <Login />
+        true ? <Component {...rest} {...props} /> : <Login />
```
morgan-sam: If you want to bypass the login page when testing, it's best to change that in the `App.js` component. So have the `"/"` route point to the component you want: `<Route path="/" exact component={Dashboard} />`

If you change ternary and if statements it's easy to forget about them and harder to find later.

client/src/pages/Friends.js
```diff
@@ -1,27 +1,99 @@
...
+  const handleFollow=(person)=>{
+    const newFollowingsList=[...followingsList]
+    newFollowingsList.push(person)
+    setFollowings(newFollowingsList)
+  }
+  const handleUnfollow=(person)=>{
+    const newFollowingsList=[...followingsList]
+    newFollowingsList.pop(person)
+    setFollowings(newFollowingsList)
+  }
```
morgan-sam: So a toggle function would check if the followingsList contains the person. If so they would remove them from the array, if not they would add that to the array (also change newFollowingsList to let as you are mutating it) :

```
  const toggleFollow = (person) => {
    let newFollowingsList = [...followingsList];
    if (followingsList.includes(person)) newFollowingsList.pop(person);
    else newFollowingsList.push(person);
    setFollowings(newFollowingsList);
  };
```

client/src/pages/Friends.js
```diff
@@ -1,27 +1,99 @@
...
+import { Box, Typography, Tabs,Tab } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
 
+import Followers from '../components/Followers'
+import Following from "components/Following";
+import rose from '../img/roseBarron.png';
+import david from '../img/davidRichardson.png'
+import christine from '../img/christineSmith.png'
+import jennifer from '../img/jenniferFordham.png'
+import rebbeca from  '../img/rebbecaSchindler.png'
```
morgan-sam: If you change these to absolute imports so:

`'../img/roseBarron.png'`

becomes:

`'img/roseBarron.png'`

it allows the files to be more modular so they aren't dependent on their own position.

client/src/components/Followers.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+    
+  }));
+const Followers =(props)=>{
+    const classes = useStyles();
+    return(
+        <List aria-label='followers list' className={classes.list}>
+            {props.followersList.map((person=>{
+              return(
+                <Fragment>
```
morgan-sam: You are getting a key error as React requires that any elements that are generated from an array, must be each supplied a unique key. This can be temporarily fixed by passing the map index to Fragment as a key, but the best way would be when we implement a unique id to each Follower to pass that instead as the key to the Fragment.

morgan-sam (commented):
The UI looks good, very similar to sketch files.

The only UI thing to change is to make sure the button text changes from 'follow' to 'unfollow' or vice versa on click, depending on the state.

client/src/components/Followers.js
```diff
@@ -0,0 +1,56 @@
...
+    unFollowButton:{
+        borderRadius: "10rem",
+        backgroundColor: "#DF1B1B",
+        color: "white",
+        width: "6rem",
+        height: "2rem",
+        fontSize:'0.6rem'
+        
+    }
+
```
ThinkTink: let's remove extra spacing

client/src/pages/Friends.js
ThinkTink: So pop doesn't actually work like this  in JavaScript, it just removes the last item from the array.

Given that we probably need to be making API calls to add or remove followers, I think in this case it might actually be better to keep it as separate functions. You could also then also pass the index of the person to be removed, which would make removing from the array more efficient.

client/src/pages/Dashboard.js
```diff
@@ -37,8 +38,12 @@ const Dashboard = () => {
         }}
       />
       <Box className={classes.bodyContainer}>
-        {selectedPage === 0 && <ShoppingLists />}
-        {selectedPage === 1 && <Friends />}
+        {/* {selectedPage === 0 && <ShoppingLists />}
+        {selectedPage === 1 && <Friends />} */}
+        <Switch>
+          <Route exact path="/dashboard" component={ShoppingLists}/> 
+          <Route exact path="/dashboard/friends" component={Friends}/>
```
Mila-ashtari: @morgan-sam @aeciorc  could you please take a look at this, I did places the routes in the dashboard. however only the first route renders and when navigating to the friends route there is a blank page

client/src/components/HeaderTabs.js
```diff
@@ -58,10 +59,16 @@ const HeaderTabs = (props) => {
...
           label={<Typography>Shopping Lists</Typography>}
+          component={Link}
+          to="/dashboard"
         />
         <StyledTab
          disableRipple 
-         label={<Typography>Friends</Typography>} />
+         label={<Typography>Friends</Typography>} 
+         component={Link}
+         to="/dashboard/friends"
```
Mila-ashtari: @aeciorc   here the tabs are converted in to link components which they work fine

client/src/pages/Friends.js
morgan-sam: That's true, the function would get too large if API calls were included as well.

What if we created a simple toggle function like this?
```
  const toggleFollow = (person) => {
      followingsList.includes(person) ? handleUnfollow(person) : handleFollow(person);
  }
```

That way we could separate the concerns of the follow vs unfollow logic, but still have the benefit of only having to pass one function to `<Followers />` & `<Following />`.

ThinkTink: yeah, I like that!

client/src/pages/Dashboard.js
morgan-sam: Friends page is working fine for me! The routing changed since this PR so the new dashboard route is `"/"`, but otherwise it is good to merge.

PR #55, opened 2020-10-28
Feature Modularize Scraper

morgan-sam:
**What it does**

The scraper was very rigid before in its design, and had a lot of unnecessary repetition. As a result it wasn't very suitable for extending the functionality to other websites. To fix that a selectors object was implemented, so all the scraping details can be programmatically retrieved from a central object.

**What to look for**

- Is variable/function naming good? The code has been shortened but it's important it's still clear what each part does.
- The selectors object is very large, so it likely needs formatting to make it more readable
- It will be great to implement another website using the selectors, but that is likely not a priority at the moment

**What not to avoid**
- **All changes are limited to the `scraper.py` file**, other changes are from the [_Feature Price History_](https://github.com/hatchways/team-apple-cider/pull/51) branch


server/api/scraper.py
```diff
@@ -3,74 +3,70 @@
 from selenium.webdriver.chrome.options import Options
 from webdriver_manager.chrome import ChromeDriverManager
 
+def loadChromeDriver():
+    chrome_options = Options()
+    chrome_options.add_argument('--no-sandbox')
+    chrome_options.add_argument('--ignore-certificate-errors')
+    chrome_options.add_argument('--headless')
+    chrome_options.add_argument('--disable-dev-shm-usage')
+    chrome_options.add_argument('--remote-debugging-port=9222')
```
ThinkTink: Maybe we can pull these out into a constant to make this cleaner. Something like:
```
CHROME_OPTIONS = ['--no-sandbox', etc...]

for option in CHROME_OPTIONS:
    chrome_options.add_argument(option)
```

server/api/scraper.py
```diff
@@ -3,74 +3,70 @@
...
+    executable_path=ChromeDriverManager().install()
+    userAgent = 'Mozilla/5.0 (X11; Linux x86_64)' +  'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.39 Safari/537.36'
+    chrome_options.add_argument("user-agent={}".format(userAgent))   
+    return webdriver.Chrome(executable_path=executable_path, options=chrome_options)    
+
+
+def string_to_int_price(price_string):
+    if price_string == None: return None
+    x = re.search(r"([0-9]+)\.([0-9]+)", price_string)
```
ThinkTink: let's rename x

server/api/scraper.py
```diff
@@ -3,74 +3,70 @@
...
+def string_to_int_price(price_string):
+    if price_string == None: return None
+    x = re.search(r"([0-9]+)\.([0-9]+)", price_string) 
+    return int(x.group(1)) * 100 + int(x.group(2))
+
+def string_availability_to_boolean(string_availability):
+    return (bool(re.search('in stock', string_availability, re.IGNORECASE)))
+
+
+selectors = {
```
ThinkTink: :+1: 
Constants are typically written in all caps

server/api/prices_handler.py
```diff
@@ -0,0 +1,53 @@
...
+        try:
+            price_entry = Price(product_id, body['price'])
+            db.session.add(price_entry)
+        except:
+            return jsonify({'error': "{}".format(e.__cause__)}), 400      
+        try:
+            db.session.commit()
+        except Exception as e:
+            return jsonify({'error': "{}".format(e.__cause__)}), 400
+        return jsonify({'response': "Added price {} to product '{}'".format(body['price'], product_id)}), 200
```
adams-liu, line 27: Maybe we can do a try statement before `body = json.loads(request.get_data())`.
It may not be necessary to have 2 try/except here.

server/models/price.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Price(db.Model):
+
+    __tablename__ = 'prices'
+
+    product = relationship('Product')
+
+    primary_id = db.Column(db.Integer, primary_key=True, unique=True, autoincrement=True)
+    product_id = db.Column(ForeignKey('product.id'), nullable=False)
+    price = db.Column(db.Integer, nullable=False)
```
adams-liu, line 16: I've removed the `nullable=False` for the products sections because I realized sometimes products may not have a price

PR #56, opened 2020-10-29
PR: Bugfix for product_handler.py

adams-liu:
**What it does**
- Just a quick bugfix, completely forgot to update the Cloudinary image after a put request.
- I removed `nullable=False` in the product model for `price `and `old_price`, because there can be instances where prices are unavailable

### DIFF
--- a/server/api/product_handler.py
+++ b/server/api/product_handler.py
@@ -1,8 +1,6 @@
 from flask import jsonify, Blueprint, request
 from models.product import Product
-from models.list import List
 from database import db
-
 from api.image_uploader import image_uploader
 from config import PRODUCT_IMG_PRESET, CLOUDINARY_NAME
 import json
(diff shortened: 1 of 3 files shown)


server/api/product_handler.py
```diff
@@ -22,23 +20,31 @@ def oneProductRequests(product_id):
...
-        if product:             
-            if "name" in req: product.name = req["name"]
-            if "old_price" in req: product.old_price = req["old_price"]
-            if "price" in req: product.price = req["price"]
-            if "url" in req: product.url = req["url"]
-            if "img_url" in req: product.img_url = req["img_url"]
+        try:
+            product = Product.query.get(int(product_id))
+            if "name" in req:
+                product.name = req["name"]
```
ThinkTink: We could make this more concise by using `product.name = req.get("name", product.name)`

adams-liu: thanks carmen!

ThinkTink (approved):

morgan-sam (approved):

PR #58, opened 2020-10-29
Integration: Scraper and Prices

morgan-sam:
**What it does**

- Scraper now makes a POST request to prices route every time it is run.
- Scraper now gets currency of item which is stored in the db and is also displayed on UI.

**What to look for**

- The scraper needs to use `url_id` as it has no way to determine the `product_id`, so the routes were set to use `url_id`. Eventually I couldn't see the point of keeping both, as `url_id` was doing the exact same thing, so I removed `product_id`. If there is a reason to keep `product_id` as well let me know (also it may make sense to keep only `url_id` but rename it to `product_id`).
- I've tested fair amount of countries that are working (USA, Canada, UK, EU), but have not exhausted all possible domains. I've found some countries, India for example, use a different URL structure.
- If there are any countries that have overlapping currency symbols, but different actual currencies.

**What to avoid**

- Only posts when scraper is run by user, no automatic async updates yet.


server/models/price.py
```diff
@@ -3,29 +3,29 @@
...
 class Price(db.Model):
 
     __tablename__ = 'prices'
 
-    product = relationship('Product')
-
     primary_id = db.Column(db.Integer, primary_key=True, unique=True, autoincrement=True)
-    product_id = db.Column(ForeignKey('product.id'), nullable=False)
-    price = db.Column(db.Integer)
+    url_id = db.Column(db.String, nullable=False)
```
ThinkTink: We need the database to be able to know how prices are connected to products.
If we're removing the product_id, maybe we should add the url_id to the product table and make that the foreign key.

server/models/price.py
```diff
@@ -3,29 +3,29 @@
...
+    price = db.Column(db.Integer)    
+    currency = db.Column(db.String)
```
adams-liu, line 19: I could be wrong, but is `currency` stored a symbol? Can we do USD, CAD, GBP?

morgan-sam: Yeah there's an issue here we need to decide on in terms of storing currency. On Amazon the domain and the currency used may not necessarily match, the user can choose a different currency. However that selection is visible at the bottom of _amazon.com_ but not on amazon in others countries .

So the only way I found you could consistently get the right currency is just scraping it straight off of the item price, which will be formatting like: "$", "£", "€", "CDN$", etc. Right now the currency is purely cosmetic, so it shows the right currency on the UI. But for instance it doesn't differentiate between Australian dollars and US dollars, they'll both be "$".

The solution I think would be to manually create a table ourselves that matches the scraped currency symbol and the domain up to a particular 3 letter international currency acronym.

server/models/price.py
morgan-sam: Ok I've connected prices to products. 

What was called url_id is just called product_id now. Its the id on Product and the foreign key on Price.

It involved changing the POST route for `/products` as you now have to provide the id when posting.

When scraping it will check if the product exists on the db, if not it will add it.

@adams-liu later today lets reconvene to make sure our changes to product aren't conflicting.

server/api/prices_handler.py
```diff
@@ -19,11 +19,12 @@ def onePriceRequests(product_id):
...
         try:
             body = json.loads(request.get_data())
-            price_entry = Price(product_id, body['price'])
+            price_entry = Price(product_id, body['price'], body['currency'])
             db.session.add(price_entry)
             db.session.commit()
-            return jsonify({'response': "Added price {} to product '{}'".format(body['price'], product_id)}), 200
+            return jsonify({'response': "Added price {} {} to product '{}'".format(body['currency'], body['price'], product_id)}), 200
         except Exception as e:
+            print(e)
```
ThinkTink:
```suggestion

```

server/api/product_handler.py
```diff
@@ -11,21 +11,39 @@
...
+            db.session.delete(product)
+            db.session.commit()
+            return jsonify({'response': "Product '{}' was successfully deleted from the database".format(product_id)}), 200
+        except Exception as e:
+            return jsonify({'error': "{}".format(e.__cause__)}), 400
     
+    if request.method == 'POST':
+        try:
+            body = json.loads(request.get_data())
+            product_item = Product(product_id, body['name'], body['currency'], body['old_price'], body['price'], body['availability'], body['url'], body['img_url'])
```
ThinkTink: We need to provide the product_id from the post body as well

server/api/scraper.py
```diff
@@ -43,15 +50,27 @@ def __init__(self, URL):
...
+        self.product_id = self.get_product_id(URL)
+
         self.old_price = self.get_parameter(driver, "old_price") 
         self.price = self.get_parameter(driver, "price") 
-        self.name = self.get_parameter(driver, "name") 
-        self.img_url = self.get_parameter(driver, "img_url") 
+        self.name = self.name = self.get_parameter(driver, "name") 
+        self.img_url = self.img_url = self.get_parameter(driver, "img_url") 
         self.availability = self.get_parameter(driver, "availability") 
+        self.currency = self.currency = self.get_parameter(driver, "currency")
```
ThinkTink: typos here?

server/models/product.py
```diff
@@ -5,17 +5,22 @@ class Product(db.Model):
...
     price = db.Column(db.Integer, nullable=False)
+    availability = db.Column(db.Boolean, nullable=False)
     url = db.Column(db.String, nullable=False)
     img_url = db.Column(db.String, nullable=False)
 
-    def __init__(self, name, old_price, price, url, img_url):
+    def __init__(self, id, name, currency, old_price, price, availability, url, img_url):
+        self.id = id
         self.name = name
+        self.currency = currency
```
adams-liu, line 20: I realized, you have a currency column for both product and prices table, I don't think this would be necessary for the prices table.

server/models/price.py
adams-liu: Now, I think about, it shouldn't be **TOO** big of a deal. I was moreso just concerned on how symbols would be read on different software platforms. I'm sure SQLAlchemy already handles that issue for us so we don't need to be concerned. It's mainly because I've worked with csv files where it has been an issue previously, but I think we should be good.

adams-liu: Also, in regards to the differentiating between different dollar symbols. I guess it would be more clear for the user to read "$60USD" but they would also be provided a link to go to the web page so they'll eventually realize the price.

server/api/product_handler.py
morgan-sam: Wouldn't the product_id be obtained from the route?
`@product_handler.route('/products/<product_id>', methods=['GET', 'DELETE', 'POST', 'PUT'])
`

ThinkTink: oh right...my bad

PR #59, opened 2020-10-30
PR: Redesign of List Models, API routes, with Authorization (Backend)

adams-liu:
**What is does**
 - Redesigned the List models, and association with products. Currently there is another model called "List_To_Products" which connects the two models. (Check image below)
- Now when performing a get/post request for certain apis, authorization is checked before the request executes.

**What to look out for**
- I'm uncertain if having `user_id` as a route parameter is safe
- In the `auth_handler ` I added a new function called `token_getter` which gives me the decoded `list_id` , thought this was better than a decorator?
- Check the try/except statements, I've just copying and pasting the same except statement everywhere. There may be a better optimized way to arrange them

**What to avoid**
- Currently, all new created list will have the item_count will be set to 0, we'll need to find the best way to increment this
- PUT and DELETE requests will be on a future PR

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,39 @@ A price history is kept for each product on the database. The API routes for the
 
 `curl -X DELETE localhost:5000/prices/product/<product_id>`
 
+### Lists
+
+The API routes for the lists are as follows:
+
+#### _GET one personal list (public or private) created by the user:_
+`curl localhost:5000/lists?user_id=<USER_ID>&list_id=<LIST_ID>`
+
+#### _GET a all list of any user (only authorized users can see private ones):_
+`curl localhost:5000/lists?user_id=<USER_ID>`
+
+#### _GET one list of any user (only authorized users can see private ones):_
+`curl localhost:5000/lists?list_id=<LIST_ID>`
+
+#### _POST(create) new list:_
+`curl -X POST localhost:5000/lists --data 
+'{
+   "name": "<NAME>",
+   "img_url": "<IMAGE URL>"
+}' --header "Content-Type: application/json"`
+
+### List to Products
+
+#### _GET all existing product_ids given an list_id (only authorized users can see private ones):_
+
+`curl localhost:5000/list-to-products/<LIST_ID>`
+
+#### _POST(add) new products into existing ones (only authorized users can do this):_
+
+`curl -X POST localhost:5000/list-to-products/<LIST_ID> --data 
+'{
+   "product_id": <PRODUCT_ID>
+}'--header "Content-Type: application/json"`
+
 ## External API Services
 
 ### Cloudinary
--- a/README.md
+++ b/README.md
@@ -91,10 +91,10 @@ The API routes for the lists are as follows:
 #### _GET one personal list (public or private) created by the user:_
 `curl localhost:5000/lists?user_id=<USER_ID>&list_id=<LIST_ID>`
 
-#### _GET a all list of any user (only authorized users can see private ones):_
+#### _GET all list of any one user (only authorized users can see private ones):_
 `curl localhost:5000/lists?user_id=<USER_ID>`
 
-#### _GET one list of any user (only authorized users can see private ones):_
+#### _GET one list of any user (public lists only):_
 `curl localhost:5000/lists?list_id=<LIST_ID>`
 
 #### _POST(create) new list:_
@@ -106,11 +106,11 @@ The API routes for the lists are as follows:
 
 ### List to Products
 
-#### _GET all existing product_ids given an list_id (only authorized users can see private ones):_
+#### _GET all existing product given an list_id (only authorized users can see ones from private lists):_
 
 `curl localhost:5000/list-to-products/<LIST_ID>`
 
-#### _POST(add) new products into existing ones (only authorized users can do this):_
+#### _POST(add) new product ids into an existing list (only authorized users can do this):_
 
 `curl -X POST localhost:5000/list-to-products/<LIST_ID> --data 
 '{
--- a/README.md
+++ b/README.md
@@ -64,6 +64,26 @@ The information is returned as a JSON object:
 }
 ```
 
+### Prices
+
+A price history is kept for each product on the database. The API routes for the pricing are as follows:
+
+#### _GET entire price history all products in database:_
+
+`curl localhost:5000/prices`
+
+#### _GET entire price history of a single product:_
+
+`curl localhost:5000/prices/product/<product_id>`
+
+#### _POST new price to price history of a single product:_
+
+`curl -X POST localhost:5000/prices/product/<product_id> --data "{\"price\": #### }" --header "Content-Type: application/json"`
+
+#### _DELETE entire price history of a single product:_
+
+`curl -X DELETE localhost:5000/prices/product/<product_id>`
+
 ## External API Services
 
 ### Cloudinary
--- a/server/api/auth_handler.py
+++ b/server/api/auth_handler.py
@@ -29,7 +29,11 @@ def token_getter():
 def login_cookie_getter(f):
     @wraps(f)
     def decorated_function(*args, **kwargs):
-        token_getter()
+        auth_token = token_getter()
+        try:
+            g.user = User.query.filter_by(id=auth_token).first()
+        except:
+            raise Exception
         return f(None)
     return decorated_function
 
--- a/server/api/auth_handler.py
+++ b/server/api/auth_handler.py
@@ -7,6 +7,7 @@ from functools import wraps
 
 auth_handler = Blueprint("auth_handler", __name__)
 
+
 def token_getter():
     try:
         auth_header = request.cookies.get("Authentication token")
@@ -24,34 +25,13 @@ def token_getter():
     return make_response(jsonify(responseObject)), 401
 
 
-
 # Wrapper that obtains and verifies an auth token in a cookie. Returns true if valid, false otherwise.
 def login_cookie_getter(f):
     @wraps(f)
     def decorated_function(*args, **kwargs):
-        try:
-            auth_header = request.cookies.get("Authentication token")
-            if auth_header:
-                auth_token = auth_header
-                auth_token = User.decode_auth_token(auth_token)
-                if not isinstance(auth_token, str):
-                    g.user = User.query.filter_by(id=auth_token).first()
-                else:
-                    raise Exception
-            else:
-                raise Exception
-        except:
-            responseObject = {
-                "status": "fail",
-                "message": "User is not logged in."
-            }
-            return make_response(jsonify(responseObject)), 401
+        token_getter()
         return f(None)
     return decorated_function
-
-
-
-
 
 
 class RegisterAPI(MethodView):
--- a/server/api/auth_handler.py
+++ b/server/api/auth_handler.py
@@ -30,11 +30,6 @@ def login_cookie_getter(f):
     @wraps(f)
     def decorated_function(*args, **kwargs):
         auth_token = token_getter()
-        try:
-            g.user = User.query.filter_by(id=auth_token).first()
-        except:
-            raise Exception
-        return f(None)
     return decorated_function
 
 
--- a/server/api/auth_handler.py
+++ b/server/api/auth_handler.py
@@ -7,6 +7,23 @@ from functools import wraps
 
 auth_handler = Blueprint("auth_handler", __name__)
 
+def token_getter():
+    try:
+        auth_header = request.cookies.get("Authentication token")
+        if auth_header:
+            auth_token = auth_header
+            auth_token = User.decode_auth_token(auth_token)
+            return auth_token
+        else:
+            raise Exception
+    except:
+        responseObject = {
+            "status": "fail",
+            "message": "User is not logged in."
+        }
+    return make_response(jsonify(responseObject)), 401
+
+
 
 # Wrapper that obtains and verifies an auth token in a cookie. Returns true if valid, false otherwise.
 def login_cookie_getter(f):
@@ -31,6 +48,10 @@ def login_cookie_getter(f):
             return make_response(jsonify(responseObject)), 401
         return f(None)
     return decorated_function
+
+
+
+
 
 
 class RegisterAPI(MethodView):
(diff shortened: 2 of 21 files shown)


ThinkTink:
So a list_to_products table is not needed. If we just add a list_id as a foreign key to the products table that will serve the same purpose

adams-liu:
> So a list_to_products table is not needed. If we just add a list_id as a foreign key to the products table that will serve the same purpose

that was going to be my original approach, however that would mean anytime a new product is added to a new list, despite being the same product, a new product entry would need to be created. This uses up extra space. And if that is the case every product of every list would need to be scraped.

server/api/auth_handler.py
```diff
@@ -8,27 +8,28 @@
...
-                    raise Exception
-            else:
-                raise Exception
-        except:
-            responseObject = {
-                "status": "fail",
-                "message": "User is not logged in."
-            }
-            return make_response(jsonify(responseObject)), 401
+        token_getter()
```
holley2032: I like what you've done with the token_getter, but there's a change that may need to be made to preserve the purpose of the login cookie_getter. I think that token_getter() should be replaced by:

```
auth_token = token_getter()
try:
    g.user = User.query.filter_by(id=auth_token).first()
except:
    raise Exception
```

This will set g.user to the current user, which is important for the User API.

holley2032 (dismissed):
I didn't test the list_handler or the list_to_product_handler, but a cursory glance at the code seems to indicate that they look fine to me. I think having user_id as a route parameter should be fine. I'm not a security expert, but the user_id is a public identifier that we don't need to keep hidden from any user. Merely having the user_id won't be enough for a hacker to do anything. The thing we have to protect is the authentication token, because that is what grants access to everything for that user.

Once the noted change is made to the auth_handler, I see no problem with that portion of code. I probably won't submit approval myself just because I am not too familiar with the list_handler and whatnot, but I trust that what you have done there makes sense.

README.md
```diff
@@ -84,6 +84,39 @@ A price history is kept for each product on the database. The API routes for the
...
+
+The API routes for the lists are as follows:
+
+#### _GET one personal list (public or private) created by the user:_
+`curl localhost:5000/lists?user_id=<USER_ID>&list_id=<LIST_ID>`
+
+#### _GET a all list of any user (only authorized users can see private ones):_
+`curl localhost:5000/lists?user_id=<USER_ID>`
+
+#### _GET one list of any user (only authorized users can see private ones):_
```
ThinkTink:
```suggestion
#### _GET one list of any user (public list only):_
```
-since this route does not check if user is authorized if only the list_id is provided

server/api/list_handler.py
```diff
@@ -1,36 +1,78 @@
...
-                    "id": item.id,
-                    "product_type": item.product_type
-                })
-        return jsonify({'list': product_types_ls}), 200
+    list_id = request.args.get('list_id', None)
+    user_id = request.args.get('user_id', None)
+    auth_token = token_getter()
+
+    if request.method == 'GET':
+        # 1|1|1 Case 1: Where list_id and user_id and auth_token provided
```
ThinkTink: `list = List.query.filter_by(id=list_id).first(` is done in every if statement, so let's pull it out and have it just once here at the top

server/api/list_handler.py
```diff
@@ -1,36 +1,78 @@
...
+        if list_id and user_id and auth_token:
+            if int(user_id) == int(auth_token):
+                try:
+                    list = List.query.filter_by(id=list_id).first()
+                except Exception as e:
+                    return jsonify({'error': "{}".format(e.__cause__)}), 400
+                return jsonify(List.query.filter_by(id=list_id).first().serialize)
```
ThinkTink: This can just be `return jsonify(list.serialize)` since we saved list to a variable.
Does this need to be in a tryexcept? What happens if that list id doesn't exist - will .first() throw and error or just return none? If it is needed, let's move the return statement inside the try/except

server/api/list_handler.py
```diff
@@ -1,36 +1,78 @@
...
+            else:
+                return jsonify({'error': "unauthorized access"}), 401
+
+        # 0|1|1 Case 2: when user_id and auth_token is provided
+        if user_id and auth_token:
```
ThinkTink: What is you aren't logged in but want to see the lists of a user, such as sharing a public profile page? Should this just be only `if user_id:` ?

server/api/list_handler.py
```diff
@@ -1,36 +1,78 @@
...
+        list_name = body['name']
+
+        # checks to see if list already exists
+        if not List.query.filter_by(user_id=list_user_id, name=list_name).first():
+            # checks to see if cloudinary works
+            try:
+                new_img_url = image_uploader(
+                    body['img_url'], PRODUCT_IMG_PRESET, CLOUDINARY_NAME)
+            except:
+                return jsonify({"error : uploading image on cloudinary"}), 400
```
ThinkTink:
```suggestion
                return jsonify({"error : Error uploading image on cloudinary"}), 400
```

README.md
```diff
@@ -84,6 +84,39 @@ A price history is kept for each product on the database. The API routes for the
...
+#### _POST(create) new list:_
+`curl -X POST localhost:5000/lists --data 
+'{
+   "name": "<NAME>",
+   "img_url": "<IMAGE URL>"
+}' --header "Content-Type: application/json"`
+
+### List to Products
+
+#### _GET all existing product_ids given an list_id (only authorized users can see private ones):_
```
ThinkTink:
```suggestion
#### _GET all existing products given an list_id (only authorized users can see ones from private lists):_
```

server/api/list_to_product_handler.py
```diff
@@ -0,0 +1,50 @@
...
+            return jsonify({'error': "{}".format(e.__cause__)}), 400
+
+    if request.method == 'POST':
+        try:
+            list_user_id = List.query.filter_by(
+                id=int(list_id)).first().user_id
+
+            body = request.get_json()
+            if list_user_id == auth_token:
+                print(int(list_id))
```
ThinkTink:
```suggestion

```

server/api/list_to_product_handler.py
```diff
@@ -0,0 +1,50 @@
...
+                list_product_connection = ListToProduct(
```
ThinkTink: should we be checking if it already exists first?

server/api/product_handler.py
```diff
@@ -11,6 +11,7 @@
 
 @product_handler.route('/products/<product_id>', methods=['GET', 'DELETE', 'PUT'])
```
ThinkTink, line 12: I think we also should be adding a ListToProduct connection when adding a product. I think you should be able to add a connection independently, like if you want to follow a product you see on a friend's list, but we'll never be adding a product but not adding it to ListToProduct, right?

README.md
```diff
@@ -84,6 +84,39 @@ A price history is kept for each product on the database. The API routes for the
...
 `curl -X DELETE localhost:5000/prices/product/<product_id>`
 
+### Lists
+
+The API routes for the lists are as follows:
+
+#### _GET one personal list (public or private) created by the user:_
+`curl localhost:5000/lists?user_id=<USER_ID>&list_id=<LIST_ID>`
+
+#### _GET a all list of any user (only authorized users can see private ones):_
```
ThinkTink:
```suggestion
#### _GET all lists of any one user (only authorized users can see private ones):_
```

README.md
```diff
@@ -84,6 +84,39 @@ A price history is kept for each product on the database. The API routes for the
...
+   "img_url": "<IMAGE URL>"
+}' --header "Content-Type: application/json"`
+
+### List to Products
+
+#### _GET all existing product_ids given an list_id (only authorized users can see private ones):_
+
+`curl localhost:5000/list-to-products/<LIST_ID>`
+
+#### _POST(add) new products into existing ones (only authorized users can do this):_
```
ThinkTink:
```suggestion
#### _POST(add) new product ids into an existing list (only authorized users can do this):_
```

server/api/list_handler.py
```diff
@@ -1,36 +1,78 @@
...
+
+            try:
+                list = List(list_user_id, list_name, new_img_url)
```
ThinkTink: We also want the privacy setting here right?

server/models/list.py
```diff
@@ -1,12 +1,32 @@
...
     __tablename__ = 'list'
 
-    id = db.Column(db.Integer, primary_key=True,autoincrement=True)
-    product_type = db.Column(db.String, unique=True, nullable=False)
-    # have an image type aswell, will implement later...
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    user_id = db.Column(db.Integer, db.ForeignKey('users.id'),nullable=False)
+    name = db.Column(db.String, nullable=False)
+    img_url = db.Column(db.String, nullable=False)
+    product_count = db.Column(db.Integer, nullable=False)
```
ThinkTink: Do we actually need this field?

server/models/list_to_product.py
```diff
@@ -0,0 +1,24 @@
+from database import db
+
+
+class ListToProduct(db.Model):
+
+    __tablename__ = 'list_to_product'
+    list_id = db.Column(db.Integer, primary_key=True, nullable=False)
+    product_id = db.Column(db.Integer,primary_key=True, nullable=False)
```
ThinkTink: We can't have two primary keys, so let's make this a foreign key

ThinkTink (commented):
Good work here :+1:

server/models/list.py
morgan-sam: It will likely be much easier to use`Array.length`on the frontend to count the amount of products. Having a product_count value in the db opens up the possibility of it going out of sync with the true count.

server/api/product_handler.py
adams-liu: Correct, I'm going to integrate it so that a POST request is made from the client side to append to the ListToProducts table and to the Products Table

server/api/auth_handler.py
```diff
@@ -8,27 +8,32 @@
...
-            else:
-                raise Exception
+            g.user = User.query.filter_by(id=auth_token).first()
         except:
-            responseObject = {
-                "status": "fail",
-                "message": "User is not logged in."
-            }
-            return make_response(jsonify(responseObject)), 401
+            raise Exception
```
ThinkTink: I don't think we want this decorator to raise an Exception, just to return a 401

server/api/list_handler.py
```diff
@@ -1,36 +1,120 @@
...
-        list_item = List(body['product_type'])
-        db.session.add(list_item)
-        try:
-            db.session.commit()
-        except Exception as e:
-            return jsonify({'error': "{}".format(e.__cause__)}), 400
-        return jsonify({'response': "{} was successfully added to the database".format(body['product_type'])}), 200
+        # client must be logged-in in order to create a list, client is only restricted to create list for their own account
+        if type(auth_token) is not int:
+            return jsonify({'error': "you must log in, to create a list"}), 400
```
ThinkTink:
```suggestion
            return jsonify({'error': "you must log in to create a list"}), 400
```

server/api/list_handler.py
```diff
@@ -1,36 +1,120 @@
...
+        # client must be logged-in in order to delete a list, client is only restricted to deleting their own lists
+        if type(auth_token) is not int:
+            return jsonify({'error': "you must logged-in, to delete the list"}), 400
+
+        else:
+            list = List.query.get(int(list_id))
+            if list.user_id == int(auth_token):
+                try:
+                    req = request.get_json()
+                    db.session.delete(list)
```
ThinkTink, line 113: We we need to update ListToProduct here as well, but that can be done in a future PR

server/api/list_to_product_handler.py
```diff
@@ -0,0 +1,73 @@
...
+            else:
+                return jsonify({"error": "user has set list to private"})
+
+        except Exception as e:
+            return jsonify({'error': "{}".format(e.__cause__)}), 400
+
+    # can only add a product to a list if logged-in and the list is owned by the user
+    if request.method == 'POST':
+        if type(auth_token) is not int:
+            return jsonify({'error': "you must log in, to add a product to a list"}), 400
```
ThinkTink:
```suggestion
            return jsonify({'error': "you must log in to add a product to a list"}), 400
```

server/api/list_to_product_handler.py
```diff
@@ -0,0 +1,73 @@
...
+            except Exception as e:
+                return jsonify({'error': "{}".format(e.__cause__)}), 400
+
+
+@list_to_product_handler.route('/list-to-products/<list_id>/<product_id>', methods=['DELETE'])
+def listToProductRequest(list_id, product_id):
+    auth_token = token_getter()
+    if request.method == 'DELETE':
+        if type(auth_token) is not int:
+            return jsonify({'error': "you must log in, to delete a product from the list"}), 400
```
ThinkTink:
```suggestion
            return jsonify({'error': "you must log in to delete a product from the list"}), 400
```

server/api/list_handler.py
```diff
@@ -1,36 +1,120 @@
...
+            body = json.loads(request.get_data())
+            list_user_id = int(auth_token)
+            list_name = body['name']
+
+            # checks to see if list already exists
+            if not List.query.filter_by(user_id=list_user_id, name=list_name).first():
+                # checks to see if cloudinary works
+                new_img_url = replace_cloudinary_image(body['img_url'])
+                try:
+                    list = List(list_user_id, list_name, new_img_url)
```
ThinkTink, line 72: privacy setting as well, right?

server/models/list.py
```diff
@@ -1,12 +1,31 @@
...
-        self.product_type = product_type
+    @property
+    def serialize(self):
+        """Return object data in easily serializable format"""
+        return {
+        "id" : self.id, 
+        "user_id" : self.user_id, 
+        "name": self.name, 
+        "img_url": self.img_url,
+        "product_count" : self.product_count,
```
ThinkTink:
```suggestion

```

ThinkTink (approved):
We'll need to be sure to update ListToProduct when deleting lists, but we can do that on the integration PR. A few other small things and it should be good to merge

PR #66, opened 2020-11-02
Feature detail list integration pt1

adams-liu:
**What it does**
- This ticket is for the first part of the integration where the client can read, add, and delete products from a list.
- Client can choose which list to add the product to on the dropdown menu.

**What to look for**
- Check if the `useEffects `are used correctly, currently it works but there maybe a lot of repeated code
- I added `ListContext`, currently there is only one function and basically checks if the `add button ` from the `AddItemPopup` has been clicked, which toggles a fetch request to update the itemcount.

**What to avoid**
- Currently, the add list from the list-pop hasn't been implement yet
- Add list needs to be implemented as well

add_product


client/src/components/ListPopup.js
```diff
@@ -90,48 +88,40 @@ const useStyles = makeStyles((theme) => ({
...
+    console.log({newProducts})
+  }
+
+  useEffect(() => {
+    getListRelations()
+  }, [listId]);
+
+  useEffect(() =>{
+    getProducts()
+  },[listToProducts])
```
aeciorc: This can be a single useEffect. I don't think you ever need to run the bottom one before the first one completes, right?

client/src/components/ListPopup.js
```diff
@@ -90,48 +88,40 @@ const useStyles = makeStyles((theme) => ({
...
+    const json = await res.json()
+    setListToProducts(json);
+    console.log({json})
+  }
+
+  const getProducts = async () => {
+    const newProducts = await Promise.all(listToProducts.map(async (relation)=>{      
+      const res = await fetch(`/products/${relation.product_id}`)
+      return res.json();
+    }));
```
aeciorc: I believe this will initially run with an empty array since the first useEffect won't have run

client/Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
+name = "pypi"
```
ThinkTink: Why do we have a pipfile in the client folder?

client/src/components/AddItem.js
```diff
@@ -72,20 +77,43 @@ const AddItem = () => {
...
 
+  useEffect(() => {
+    getLists();
+  }, []);
+
+
+  useEffect(() => {
+    getListId();
+  }, [selectedList]);
+
```
ThinkTink: Rather than using a useEffect, we just want a function like changeList that get calls when you pick a different one on the drop down. Also, rather than passing the id  of the list, that function should be passed the index of the list being chosen. This will not only be more efficient but will also handle the use case where a user happens to name two lists the same thing

client/src/components/AddItemPopup.js
```diff
@@ -50,8 +51,25 @@ const useStyles = makeStyles((theme) => ({
...
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ product_id: item.product_id }),
+    });
+    const json = await res.json();
+    return json;
+  };
+
+  const addButtonClick = () => {
+    addProductToList()
```
ThinkTink: don't forget semicolons! Do you have prettier set up?

client/src/components/ListCard.js
```diff
@@ -56,6 +61,17 @@ const ListCard = (props) => {
...
+  const getItemCount = async () => {
+    const res = await fetch(`/list-to-products/${listId}`)
+    const json = await res.json();
+    setItemCount(json.length)
+  };
+
+  useEffect(() => {
+    getItemCount()
+  }, [listOpen,listChange]);
+
```
ThinkTink: You are making too many API calls. We most likely want all of the relevant data fetched once, when the page loads, at the top level component that needs any of that data (dashboard?). So here for example, we should have access to the list details and when the list changes we can just set itemCount to the length of the new list that's chosen rather than make another API call.

client/src/components/ListPopup.js
```diff
@@ -90,48 +88,38 @@ const useStyles = makeStyles((theme) => ({
...
   const classes = useStyles();
   const handleClose = () => changeListOpen();
+  const [listToProducts, setListToProducts] = useState([]); // list-to-products table {list_id, product_id}
+  const [products, setProducts] = useState([]); // products table {product_id, img_url, price}
+
+  const getListRelations = async () => {
+    const res = await fetch(`list-to-products/${listId}`)
+    const json = await res.json()
+    setListToProducts(json);
+    console.log(json)
```
ThinkTink:
```suggestion

```

client/src/components/ListPopup.js
```diff
@@ -90,48 +88,38 @@ const useStyles = makeStyles((theme) => ({
...
+  }
+
+  const getProducts = async () => {
+    const newProducts = await Promise.all(listToProducts.map(async (relation)=>{      
+      const res = await fetch(`/products/${relation.product_id}`)
+      console.log(res)
```
ThinkTink:
```suggestion

```

client/src/components/ListPopup.js
```diff
@@ -90,48 +88,38 @@ const useStyles = makeStyles((theme) => ({
...
+    setProducts(newProducts)
+  }
+
+  useEffect(() => {
+    getListRelations()
+  }, [listId,listOpen]);
+
+  useEffect(() =>{
+    getProducts()
+  },[listToProducts])
```
ThinkTink: same thoughts here: 
1) we don't want to make API calls every time these change, and
2) rather than using useEffect we want a named function like handleListOpen that gets called at the relevant times.
-but - something like useEffect with listId to update local settings every time the component receives a new listId would make sense

client/src/components/ListsDisplay.js
```diff
@@ -47,29 +44,47 @@ const useStyles = makeStyles((theme) => ({
...
+    setLists(json);
+  }
+
+  useEffect(() => {
+    getUserId();
+  }, []);
+
+    useEffect(() => {
+    getLists();
+  }, [userId]);
```
ThinkTink: Something like this could be combined into a single useEffect that does both once on page load. but it looks like you added getUserId to the userContext, so you can just access it from there rather than making another API call here.

client/src/components/ProductCard.js
```diff
@@ -30,16 +30,25 @@ const useStyles = makeStyles((theme) => ({
...
 
-const ProductCard = ({item}) => {
+const ProductCard = ({item, listId,getListRelations}) => {
   const classes = useStyles();
+
+  const removeButtonClick = async () => {
+      const res = await fetch(`list-to-products/${listId}/${item.id}`,{
+      method:'DELETE'
+      })
+      console.log(await res.json())
```
ThinkTink:
```suggestion

```

client/src/components/ProductCard.js
```diff
@@ -30,16 +30,25 @@ const useStyles = makeStyles((theme) => ({
...
+      getListRelations()
```
ThinkTink: So, while you can do something like this, I think it would be better to 1) return the updated data from the backend (probably not ideal in this case), or 2) just update the local copy of the data

client/src/contexts/UserContext.js
```diff
@@ -34,15 +78,39 @@ export function UserStore(props) {
...
+        }
+      })
+      .catch((error) => {
+        console.error("Error:", error);
+        setUser(false);
+      });
   };
 
+  const [userId, setUserId] = useState('');
+  const getUserId = async () =>{
```
ThinkTink: We should do this in the checkCookie function that calls /auth/status above

client/src/contexts/ListContext.js
```diff
@@ -0,0 +1,22 @@
+import React, { useState } from "react";
+const ListContext = React.createContext({});
+
+export function ListStore(props) {
+const [listChange, setListChange] = useState(true);
+
+const listToggle = () => {
+    setListChange(!listChange);
+}
```
ThinkTink, line 9: What is the purpose of this? I don't see it being used elsewhere except one call of this function.
If we want to use context, it should be for all of the list data and API calls, but I think we can do without the context at all since every component that needs access to list data should have a shared parent component of the dashboard

ThinkTink (changes requested):
Two things:
1) can you merge dev into this branch so we can see just your changes?
2) I think we need a major rethink of what API calls happen where

client/src/components/ProductCard.js
adams-liu: @Mila-ashtari, can you look how to does this

client/src/components/AddItem.js
```diff
@@ -86,42 +101,54 @@ const AddItem = () => {
...
-          disableUnderline
-        >
-          <MenuItem value="none" disabled>
-            Select List
-          </MenuItem>
-          {demoListsArray.map((listName, i) => (
-            <MenuItem key={i} value={listName}>
-              {listName}
+
+      <form></form>
```
ThinkTink:
```suggestion

```

client/src/components/AddItemPopup.js
```diff
@@ -50,8 +51,25 @@ const useStyles = makeStyles((theme) => ({
...
+  const listToggle = useContext(ListContext).listToggle;
+
+  const addProductToList = async () => {
+    const res = await fetch(`/list-to-products/${listId}`, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ product_id: item.product_id }),
+    });
+    const json = await res.json();
+    return json;
```
ThinkTink, line 65: if we're not doing anything with this return value we don't need to return anything

client/src/components/ListCard.js
```diff
@@ -43,18 +44,53 @@ const useStyles = makeStyles((theme) => ({
...
+  }, [listToProducts]);
+
+  useEffect(() => {
+    getProductIds()
+  }, [listChange]);
+
+
+  // const changeAddProductOpen = () => {
+  //   setAddProductOpen((previous) => !previous);
+  // };
```
ThinkTink, line 93:
```suggestion
```

server/api/product_handler.py
```diff
@@ -11,12 +11,12 @@
 
 @product_handler.route('/products/<product_id>', methods=['GET', 'DELETE', 'POST', 'PUT'])
 def oneProductRequests(product_id):
-   
     if request.method == 'GET':
         try:
             product = Product.query.get(product_id)
             return jsonify(product.serialize), 200
         except Exception as e:
+            print(e)
```
ThinkTink:
```suggestion

```

server/api/product_handler.py
```diff
@@ -37,6 +37,7 @@ def oneProductRequests(product_id):
             db.session.commit()
             return jsonify({'response': "{} was successfully added to the database".format(body['name'])}), 200
         except Exception as e:
+            print(e)
```
ThinkTink:
```suggestion

```

client/src/components/ListsDisplay.js
```diff
@@ -52,22 +50,23 @@ const useStyles = makeStyles((theme) => ({
...
+  const getLists = async () =>{
+    const res = await fetch(`/lists?user_id=${userId}`)
+    const json = await res.json()
+    setLists(json);
+  }
+
+    useEffect(() => {
+    getLists();
+  }, [userId]);
+
```
ThinkTink, line 69: So this one is being done in two places, so let's move it to either the parent component or to the list context

ThinkTink (approved):
We might eventually want to do more rearranging on what API calls are where in the future. For now though, let's make the getLists call happen in only one place then it will be good to merge.

client/src/components/ProductCard.js
Mila-ashtari: yes

PR #67, opened 2020-11-02
Bugfix: Scraper High Prices

morgan-sam:
**What it does**
- Fixes bug where items prices with more than 3 digits would scrape in correctly
    - Scraper now removes all non numeric units and returns the parsed int as the number of cents, instead of splitting the string based on commas/dots
    - Note: currently (& previously) only works with currencies that have a fractional unit of 1/100 (i.e. [Japan](https://www.amazon.co.jp) or [India](https://www.amazon.in/) Amazon stores will both display incorrectly)
- Also re-fixes incorrect white-space in currency display

### DIFF
--- a/client/src/components/ItemDisplay.js
+++ b/client/src/components/ItemDisplay.js
@@ -56,10 +56,24 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+const reverseString = (string) => string.split("").reverse().join("");
+
+const addCommasToDollars = (dollars) => {
+  let dollarsString = String(dollars);
+  dollarsString = reverseString(dollarsString);
+  dollarsString = dollarsString.replace(/(.{3})/g, "$1,");
+  dollarsString = reverseString(dollarsString);
+  return dollarsString.startsWith(",")
+    ? dollarsString.substring(1)
+    : dollarsString;
+};
+
 const centsToDollarsDisplay = (inputCents) => {
   const dollars = String(Math.floor(inputCents / 100));
   const cents = String(inputCents % 100);
-  return `${dollars}.${cents.length > 1 ? cents : "0" + cents}`;
+  const dollarsString = addCommasToDollars(dollars);
+  const centsString = cents.length > 1 ? cents : "0" + cents;
+  return `${dollarsString}.${centsString}`;
 };
 
 const ItemDisplay = (props) => {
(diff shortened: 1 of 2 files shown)


client/src/components/ItemDisplay.js
```diff
@@ -56,10 +57,24 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+const reverseString = (string) => string.split("").reverse().join("");
+
+const addCommasToDollars = (dollars) => {
+  let dollarsString = String(dollars);
+  dollarsString = reverseString(dollarsString);
+  dollarsString = dollarsString.replace(/(.{3})/g, "$1,");
```
adams-liu: what if it's over $2000? Does it this change to `$1`

client/src/components/ItemDisplay.js
```diff
@@ -56,10 +57,24 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+const reverseString = (string) => string.split("").reverse().join("");
```
adams-liu: This is kind of clever and kind of hacky, I don't know how I feel about it. lol

client/src/components/ItemDisplay.js
morgan-sam: $1 is just regex for first argument, the comma that comes after is what is being inserted

client/src/components/ItemDisplay.js
morgan-sam: I can give an example why it includes reverseString. Say you get have 1000000 as an input and you want to add commas.

We want to add commas every 3 numbers. **But every three numbers from right to left**. Replace works left to right. So it reverses the string, adds the commas, then re-reverses the string:
```
> 1000000 - Input
> 0000001 - Reverse string
> 000,000,1 - Insert commas
> 1,000,000 - Reverse string
```
Sometimes it adds a comma to the very end that's why it removes it in the return.

Having said all that, it's probably not the most efficient way of doing it. From what I understand `Array.reverse()` is a reasonably expensive function to call. If there is a way to insert the commas right to left without reversing strings lets go with that way.

adams-liu (approved):

client/src/components/ItemDisplay.js
```diff
@@ -56,10 +57,20 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+// Adds a comma once every 3 digits right to left
+const addCommasToDollars = (dollars) => {
+  let dollarsArray = String(dollars).split("");
```
ThinkTink: You use String() on line 69 when defining dollars, so it shouldn't be necessary here

ThinkTink (approved):

PR #77, opened 2020-11-05
PR: Created Followers table and API requests

adams-liu:
**What it does**
- Created an association table for user model to create followed and following relationship between user instances.
- Created api requests to make calls to the followers table

**What to avoid**
- the following/followed route has to be created to be integrated with FE

image


server/api/follower_handler.py
```diff
@@ -0,0 +1,66 @@
...
+    auth_token = token_getter()
+    try:
+        user = User.query.get(user_id)
+        followings = user.followed.all()
+        return jsonify([user.serialize for user in followings]), 200
+    except Exception as e:
+        return jsonify({'error': "{}".format(e.__cause__)}), 400
+
+
+@follower_handler.route('/follow/<user_id>', methods=['GET'])
```
ThinkTink: We don't want to be making changes on a GET request, so let's make this a POST

server/api/follower_handler.py
```diff
@@ -0,0 +1,66 @@
...
+                user = User.query.get(auth_token)
+                follow_user = User.query.get(user_id)
+                user.follow(follow_user)
+                db.session.commit()
+            except Exception as e:
+                return jsonify({'error': "{}".format(e.__cause__)}), 400
+            return jsonify({'response': "you've successfully followed user {}".format(user_id)}), 400
+
+
+@follower_handler.route('/unfollow/<user_id>', methods=['GET'])
```
ThinkTink: same here, maybe DELETE instead of GET.

server/api/follower_handler.py
```diff
@@ -0,0 +1,66 @@
...
+            return jsonify({"error": "you can't unfollow yourself"}), 400
+        else:
+            try:
+                user = User.query.get(auth_token)
+                unfollow_user = User.query.get(user_id)
+                user.unfollow(unfollow_user)
+                db.session.commit()
+            except Exception as e:
+                return jsonify({'error': "{}".format(e.__cause__)}), 400
+            return jsonify({'response': "you've successfully unfollowed user {}".format(user_id)}), 400
```
ThinkTink: We could also simplify this, and just have a GET, POST, and DELETE to a single /followers/user_id route

server/models/user.py
```diff
@@ -1,25 +1,51 @@
...
+    def follow(self, user):
+        if not self.is_following(user):
+            self.followed.append(user)
+
+    def unfollow(self, user):
+        if self.is_following(user):
+            self.followed.remove(user)
+
+    def is_following(self, user):
+
```
ThinkTink:
```suggestion
```

ThinkTink (approved):

server/api/follower_handler.py
```diff
@@ -0,0 +1,66 @@
+from flask import jsonify, Blueprint
+from models.user import User
+from database import db
+from api.auth_handler import token_getter
+
+follower_handler = Blueprint('follower_handler', __name__)
+
+
+@follower_handler.route('/followers/<user_id>', methods=['GET'])
```
aeciorc: This should be `GET /followers`. You don't need the user_id since it's in the token

server/api/follower_handler.py
```diff
@@ -0,0 +1,66 @@
...
+    auth_token = token_getter()
+    try:
+        user = User.query.get(user_id)
+        followers = user.followers.all()
+        return jsonify([user.serialize for user in followers]), 200
+    except Exception as e:
+        return jsonify({'error': "{}".format(e.__cause__)}), 400
+
+
+@follower_handler.route('/followings/<user_id>', methods=['GET'])
```
aeciorc: same here

server/api/follower_handler.py
```diff
@@ -0,0 +1,66 @@
...
+from models.user import User
+from database import db
+from api.auth_handler import token_getter
+
+follower_handler = Blueprint('follower_handler', __name__)
+
+
+@follower_handler.route('/followers/<user_id>', methods=['GET'])
+def getFollowers(user_id):
+    auth_token = token_getter()
```
aeciorc, line 11: Any reason why you guys don't use the login_required decorator and instead call token_getter in each request?

PR #78, opened 2020-11-06
Feature add product integration

holley2032:
--What it does--
Integrates the front-end add product from list feature with the back-end.

--What it avoids--
It isn't perfect quite yet, there's plenty of cleanup on the code side to trim things no longer needed, and there are a few tweaks I want to make in order to make the UX better. That being said, since it's late and I don't know what type of time I'll have tomorrow, I wanted to push a functioning integration that does the job well (I think) and then I will hopefully try to make it better tomorrow if I have time.

### DIFF
--- a/client/src/components/AddProduct.js
+++ b/client/src/components/AddProduct.js
@@ -1,5 +1,4 @@
-import React, { useState } from "react";
-import { makeStyles } from "@material-ui/core/styles";
+import React, { useState, useEffect, useContext } from "react";
 import {
   Box,
   Modal,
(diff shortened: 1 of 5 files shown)


client/src/components/AddProduct.js
```diff
@@ -22,15 +24,67 @@ const AddProduct = (props) => {
...
   } = props;
   const classes = useStyles();
-  const [url, setUrl] = useState("");
+  const [inputLink, setInputLink] = useState("");
   const [list, setList] = useState("");
   const [errors, setErrors] = useState({});
+  const userId = useContext(UserContext).userId;
+  const [item, setItem] = useState({});
+  const [selectedListIndex, setSelectedListIndex] = useState(0);
+  const [listId, setListId] = useState("");
```
ThinkTink, line 36: Do we need list, listId, and selectedListIndex? Can we store just the index and use that to access the list and listId when needed?

ThinkTink (approved):
It could use a little cleanup, but at this point I think it's fine to merge it!

client/src/components/AddProduct.js
holley2032: Quite possibly, it could definitely be a lot cleaner in general, I left that for later because I wanted to get something functioning and good to go on the dev branch, but there is some cleanup in general and some spare parts left over from various testing and trying to get it to work that should be taken out.

PR #79, opened 2020-11-06
PR: Deleting Lists

adams-liu:
**What it does** 
- Created a popup that asks the user if they want to permanently delete the list
- Made delete request to delete all products within a list, when list is deleted

**What to look for**
- Coding conventions in React
- I created another toggle in `ListContext`, don't know if there is a better approach to automatically refresh

delete_list


client/src/components/AddItem.js
```diff
@@ -52,14 +54,22 @@ const useStyles = makeStyles((theme) => ({
...
+  
 }));
 
 const AddItem = () => {
   const userId = useContext(UserContext).userId;
   const [inputLink, setInputLink] = useState("");
   const [item, setItem] = useState({});
   const [popupOpen, setPopupOpen] = useState(false);
-  const [selectedListIndex, setSelectedListIndex] = useState(0);
+  const [selectedListIndex, setSelectedListIndex] = useState("");
```
aeciorc, line 72: Shouldn't this be a number?

server/api/list_to_product_handler.py
```diff
@@ -53,6 +53,23 @@ def listToProductsRequest(list_id):
             except Exception as e:
                 return jsonify({'error': "{}".format(e.__cause__)}), 400
 
+    if request.method == 'DELETE':
+        if type(auth_token) is not int:
+            return jsonify({'error': "you must log in to add a product to a list"}), 400
```
aeciorc: Isn't this a delete?

client/src/components/ListSettingsPopup.js
```diff
@@ -0,0 +1,54 @@
...
+  settingsClose,
+}) {
+  const listDeleteToggle = useContext(ListContext).listDeleteToggle;
+
+  const handleDeleteList = async () => {
+    await fetch(`/list-to-products/${listId}`, {
+      method: "DELETE",
+    });
+    await fetch(`/lists?list_id=${listId}`, {
+      method: "DELETE",
```
aeciorc: Why does it need two api calls to delete?

client/src/components/ListsDisplay.js
```diff
@@ -65,7 +67,7 @@ const ListsDisplay = (props) => {
 
     useEffect(() => {
     getLists();
-  }, [userId]);
+  }, [listDelete]);
```
aeciorc, line 70: I see what you did, but the better approach would have been for `const [lists, setLists] = useState([]);` to be in the context. Then instead of re-fetching when you delete a list, you could just remove the deleted list from the state directly

aeciorc (commented):
There are a few issues, but since time is tight, you can merge it if it works without any bugs

client/src/components/AddItem.js
adams-liu: this makes it so the default text is "Select List" rather than the first item on the list.

PR #80, opened 2020-11-06
Adding  lists

Mila-ashtari:
--what it does---
list are created with and image and a title and added to the server .
--what to look for--
the created lists are not shown upon creation . listDisplay component is not updated upon list creation.


adams-liu:
Can you please update your current branch from dev, test it, and then merge it? Thanks!

morgan-sam (approved):
Looks good, you don't have to logout you can see the list after refreshing the page.
Let's merge as it is, and fix the update on a separate branch.

PR #83, opened 2020-11-06
Feature friends integration

morgan-sam:
**What it does**

- Rewrites friend routes & tabs reset on refresh
- Adds user profiles with follow/following counts
- Integrates follower backend to allow following via friends list/profiles
- Follow back info on profile via `/follower_relation` route
- Adds explore page to view all other profiles
- Profile photo upload
- Fixed new list UI
- Fixed list popup UI
- Fixed Add new item pop from the list popup UI

**What next**

- Can still open add item popup without selecting a list
- Need to refresh to see new list added


client/src/components/ProfileMenu.js
```diff
@@ -40,9 +44,73 @@ const ProfileMenu = (props) => {
     props.history.push("/");
   };
 
+  const uploadProfilePicture = async () => {
+    const auth = await (await fetch("/auth/status")).json();
+    const userId = auth.data.user_id;
```
ThinkTink, line 49: I think this is already available from the updated user Context that Adams added

client/src/components/ProfileMenu.js
```diff
@@ -40,9 +44,73 @@ const ProfileMenu = (props) => {
...
+    await fetch(`/profiles/${userId}`, {
+      method: "PUT",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ photo: imageURL }),
+    });
+  };
+
+  const retrieveUserIcon = async () => {
+    const auth = await (await fetch("/auth/status")).json();
```
ThinkTink, line 58: We definitely don't want to be doing this twice in one component

ThinkTink (approved):
This could benefit from some cleanup, but I don't see any red flags, so at this point let's just merge it!

PR #93, opened 2020-11-17
Feature: Default Lists

morgan-sam:
**What it does**

- Displays 3 example lists with pictures when user has no lists
- Clicking on a picture creates that list for that user
- Also fixed delete list functionality (incorrect context function caused app to crash)

**Current issue**

- When a user has lists and the page is refreshed it will very briefly show the default lists before switching to the user lists


adams-liu (approved):
I made a couple of changes to the product toggling, for some reason when I added a product the lists did not refresh. I think it would be good if we merge this branch now and make other branches to rest of the issues.

